PR #48, opened 2025-02-12
Minimal pyproject.toml file

### LINKED ISSUE #28: Create `pyproject.toml` file for Hatchling and tests build
Create and add a `pyproject.toml` file that uses Hatchling: https://hatch.pypa.io/latest/

manuGil:
Added a minimal example for the project metadata which uses Hatchling a build tools. The build passes without error. 
Closes #28

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,8 @@
 out/
 .DS_Store
+
+
+# Python
+venv/
+__pycache__/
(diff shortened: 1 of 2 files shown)


jreyns:
You can add me to the maintainers, as I am also the 'owner' of the pypi project

PR #69, opened 2025-02-19
Create conda environment file. Fixes #63

### LINKED ISSUE #63: Create environment file
Create a conda environment file with the current dependencies.

aysunrhn:
Added a new file `environment.yml` with the dependencies we have identified so far.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,11 @@
+name: sedtrails
+
+dependencies:
+  - python=3.13
+  - numpy
+  - pandas
+  - matplotlib
+  - xarray
+  - typer
+  - pytest
+  - flake8
--- a/environment.yml
+++ b/environment.yml
@@ -11,4 +11,4 @@ dependencies:
   - flake8
   - sphinx=8.2.0
   - myst-parser=4.0.1
-  - sphinx-rtd-theme=3.0.2
+  - sphinx-rtd-theme
--- a/environment.yml
+++ b/environment.yml
@@ -9,3 +9,6 @@ dependencies:
   - typer
   - pytest
   - flake8
+  - sphinx=8.2.0
+  - myst-parser=4.0.1
+  - sphinx-rtd-theme=3.0.2


environment.yml
```diff
@@ -0,0 +1,14 @@
+name: sedtrails
+
+dependencies:
+  - python=3.13
```
manuGil, line 4: We agree to develop in Python 3.13. However, I suggest not to be that strict here. Can we change this to use version 3.11 and newer?

manuGil: @aysunrhn  I forgot this is conda, and python will be included in the installation. Keep the current version and ignore my comment please.

environment.yml
```diff
@@ -0,0 +1,14 @@
...
+  - numpy
+  - pandas
+  - matplotlib
+  - xarray
+  - typer
+  - pytest
+  - flake8
+  - sphinx=8.2.0
+  - myst-parser=4.0.1
+  - sphinx-rtd-theme=3.0.2
```
manuGil: vertion 3.0.2 for this package is not available in conda forge. Therefore installation fails. Please remove the version for sphinx-rtd-theme

manuGil (changes requested):
Please remove the version of the last package, the installation fails

environment.yml
aysunrhn: OK. Removed the version and it installs fine now

environment.yml
aysunrhn: We can still include the python version in the environment file, though. Do you suggest we remove the python dependency from the file and include it in the command when creating the conda environment instead?

manuGil: No, keep the Python line as it is

environment.yml
manuGil: @aysunrhn  Did you pushed the change without a version number? I do not see it in the commits

aysunrhn: I did it just now.

manuGil: Thank you

manuGil (approved):
All conversations are resolved

PR #72, opened 2025-02-19
Plugin system for format and physics converters

### LINKED ISSUE #65: Understand how plugins are managed in Python
https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/#

manuGil:
This PR implements a plugin system for the physics converter, and provide a start to do the same on the format converter. 

- The system implemented here relies on naming conventions for the classes associated with the plugins, and the `importlib` package. The latter can import modules dynamically based on their names.  I assumed there is at leas one per plugin per type of particle, ex. sand, mud, etc.
- The `BasePhysicsPlugin` is an abstract class that define the interface (methods) that every plugin must implement or use. This enforces consistency between plugins. Notice that right now I lack the knowledge to know which methods are common to all physics converter; therefore I only added the obvious ('convert').
- Plugins must inherit from the `BasePhysicsPlugin`, implement the methods required to do a conversion, and be placed in the same directory.
- Common logic to all physics conversions must be added to `physics_converter.py` which is responsible to use the correct plugin based on the modeler's choice (in config file). I included a simple example. 
- The previous structure of the `converters` package has changed.

Closes #65

### DIFF
--- a/src/sedtrails/converters/__init__.py
+++ b/src/sedtrails/converters/__init__.py
@@ -1,1 +1,7 @@
-#populate this file
+from .plugins.physics.plugin import BasePhysicsPlugin
+from .plugins.format.plugin import BaseFormatPlugin
+
+__all__ = [
+            "BasePhysicsPlugin",
+            "BaseFormatPlugin"
+           ]
--- a/src/sedtrails/converters/__init__.py
+++ b/src/sedtrails/converters/__init__.py
@@ -1,7 +1,0 @@
-from .plugins.physics.plugin import BasePhysicsPlugin
-from .plugins.format.plugin import BaseFormatPlugin
-
-__all__ = [
-            "BasePhysicsPlugin",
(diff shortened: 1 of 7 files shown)


npannozzo (approved):
Hi Manuel, I will not comment on the code itself because my knowledge of the python language and structure is still limited. I will limit my review to the logics followed to implement the plugins. I think you managed to grasp the logics behind what we discussed and it seems to follow the correct flow, so I will approve from my side.

manuGil:
> Hi Manuel, I will not comment on the code itself because my knowledge of the python language and structure is still limited. I will limit my review to the logics followed to implement the plugins. I think you managed to grasp the logics behind what we discussed and it seems to follow the correct flow, so I will approve from my side.

Thaks @npannozzo   Your insight on the logic will be perfect, we can fine tune the implementation along the way.

Issue #56, opened 2025-02-19
Implement Physics Library with first physics methods


manuGil:
We need to implement a library for methods for the Physics converter. The library is a Python package that contains Python files. I suggest to place a method per file. Each method should have unit test for [Pytest](https://docs.pytest.org/en/stable/getting-started.html)

## Where to start
We can start with the Soulsby et al (2011) method, which would be a method in the physics library found in the Transport Converter: https://github.com/sedtrails/sedtrails/blob/dev/img/v5/SedtrailComponentsTransportConverter.png

Some of this is already done in the old matlab code here, but we may need to redo it, particularly if the data structure changes:
[sand_soulsby2011_2d.txt](https://github.com/user-attachments/files/18851629/sand_soulsby2011_2d.txt)

Right now a lot of these are built-in functions within the sand_soulsby2011_2d.m function, but many of them are globally useful if we expand our library of sediment transport functions, so they should probably be brought out.  Also, we should probably redo these in an appropriately Pythonic format- classes? I'm not sure how this works.

Grain properties
- Dstar
- theta (Shields number)
- w_s (settling velocity)
- theta_critical_A (with hiding and exposure)

Soulsby
- a
- b
- R_susp
- R_bedload
- F
- P
- Ugr

Other
- Rouse number

manuGil:
Please notice that a way to implement part of this following a plugin-system approach is suggested in #72

manuGil:
This was completed by implementing a plugin system for physic conversion methods

PR #75, opened 2025-02-21
Part R of the soulsby2011_2d

monica-aguil:
My part of the code to merge at some point with the work by Natascia and Hassan. I created dummy matrices at the beginning since I don't have a sample S matrix. #56 (related to it, though not finished).


manuGil:
Belong to physics converter, a method to convert physical properties of input data

manuGil (changes requested):
Could you please put this code in a function? or use more functions if needed

manuGil:
This has been moved to new files. It will be closed.

PR #76, opened 2025-02-21
ST-44: add RK4 interpolator for particle advection

### LINKED ISSUE #44: Code up RK4 integrator
Write a python implementation of the RK integrator of SedTRAILS

jreyns:
Add ParticlePositionCalculator class and unit tests

Closes #44

### DIFF
--- a/src/sedtrails/interpolate/particlepositioncalculator.py
+++ b/src/sedtrails/interpolate/particlepositioncalculator.py
@@ -1,0 +1,300 @@
+"""
+ParticlePositionCalculator Class
+
+This class performs particle advection over an unstructured grid using a
+4‑stage Runge–Kutta integration scheme. It supports barycentric interpolation
+of grid fields (e.g., velocities) via a matplotlib Triangulation, with an option
+to use multiprocessing for very large particle sets.
+
+Usage:
+    1. Initialize with grid nodes, velocities, and face‐node connectivity.
+    2. Call update_particles() with particle positions, time step, and diffusion.
+    
+Dependencies: numpy, matplotlib, multiprocessing, typing, math
+"""
+
+import numpy as np
+from numpy.typing import NDArray
+import matplotlib.tri as mtri
+from math import pi
+import multiprocessing as mp
+from typing import Callable, Tuple
+
+
+class ParticlePositionCalculator:
+    def __init__(self, grid_x: NDArray[np.float64], grid_y: NDArray[np.float64], grid_u: NDArray[np.float64], grid_v: NDArray[np.float64], igeo: int=0):
+        """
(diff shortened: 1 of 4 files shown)


sgpearson17:
Very cool, Johan! I just started looking through this and I think I will need you to walk me through a few parts (should I be able to run ```test_particleinterpolator.py```, but overall I think it is great to see what you've done.  I like all the different cases that you test. A few first questions:
1. From a workflow perspective (more generally for the rest of the project too), if I have comments on e.g. a particular line, how should I best do that? Here? Add in-line comments to the code?
2. Right now in ```particlepositioncalculator.py``` you have a purely random walk diffusion at the very end of the code- is it possible that we could break this out into some sort of external function instead? It would be nice if we can choose from a (future) library of diffusion schemes, even if the default is just this approach that you've implemented.

jreyns:
Hi @sgpearson17 

- You can just add comments to a line of code
- Yes, we could add diffusion functions by passing them as function handles. We need to think how to set up the code for the diffusion function library

sgpearson17:
I looked through the code and think it makes sense. I was also able to successfully run the unit tests.

src/sedtrails/interpolate/particlepositioncalculator.py
```diff
@@ -0,0 +1,300 @@
...
+
+        # Stage 4.
+        up4 = get_interp(grid_u_adj, x1c, y1c)
+        vp4 = get_interp(grid_v_adj, x1c, y1c)
+
+        # Combine stages (RK4 integration).
+        x_new = x0 + dt / 6.0 * (up1 + 2.0 * up2 + 2.0 * up3 + up4)
+        y_new = y0 + dt / 6.0 * (vp1 + 2.0 * vp2 + 2.0 * vp3 + vp4)
+
+        # Add random diffusion if requested.
```
sgpearson17, line 282: Ideally we should break this out into a separate function of sorts so that we can have different diffusion methods

sgpearson17 (approved):
I reviewed the code here and was also able to run the unit tests successfully (huzzah!). My only request would be to make a separate function for diffusion but perhaps I shall leave this as-is and create a separate issue for that.

src/sedtrails/interpolate/particlepositioncalculator.py
jreyns: like i said, we can pass a function handle to a diffusion library member function at a later point

sgpearson17: Yup! I was also just experimenting with the review/commenting workflow and thought I'd mention it. Not a high priority for now though. I'm just stoked that we have a working interpolator!

Issue #30, opened 2025-02-06
Define a list of configuration options for the Configuration file


manuGil:
We need to define a list of the parameters/configurations that are necessary to run simulations in SedTRAILS.
We also need to identify which ones are  compulsory (**must be provided by the user**) and which ones are optional (we can apply defaults to).

We can document that in a table like this:

|Parameter | Type  |  Custom values? | Optional User (yes/no) |  Default value |
|-----------------------|----------------------|---------------|-------------|-----------|
| timestep | interger | no | yes  | 100 seconds|

- Type: data types such as integers, strings, URL, etc. 
- Custom values: it might be that for some parameters we want to impose fix values, such as 'high, low, etc.'
- Default value: if a parameter is optional, we should assign a default.

@sgpearson17  and @npannozzo 
I think you are the two people who has the most experience on this one. Maybe you can take on this issue.
Are units something important to include as part of the parameters?

sgpearson17:
I have a whole list of parameters and default settings from the existing code (~300 lines, see below). Units are indeed important, at least for the physical parameters. I can help format this as a table next week:
```
% Preprocess hydrodynamics etc?
S.preprocess = 1; % 1 = pre-process, 0 = use existing files with default names
S.computePathways=1; % (re)run everything = 1, load existing pathways = 0;
S.plotPathways=0; % plot pathways = 1, don't plot pathways = 0

S.inputModelType = 'dfm'; % 'dfm' = D-Flow FM, 'd3d4' = Delft3D-4, 'xbeach' = xbeach, 'aeolis' = aeolis, etc...
S.supportedModelTypes = {'dfm','d3d4'};

% input/output directories
S.inputDir=['']; % parent directory containing folders with D-Flow FM model output
S.outputDir=['\sedtrails\in\sedtrails_run_v001\']; % directory for processed DFM model output (i.e., sedtrails input)
S.sharedDir = '\sedtrails\shared\';  % general project setup directory for sedtrails with common files 
S.dataDir = []; % directory containing validation data

% batching
S.nRuns = 1; % number of runs

S.displayInputMetadata = 0; % display all metadata

% specify subsets of data in space/time/sediment fractions
S.useFullModelExtents = 1;      % use full model domain (1) or only a subset (0)?
S.subset_x = [];%[35000 52000]; % limits of model domain in x-direction
S.subset_y = [];%[1000 21000];  % limits of model domain in y-direction
[...]

PR #78, opened 2025-02-26
ST-31: Add input validation

### LINKED ISSUE #31: Create YAML schema for configurations in SedTRAILS.
Once we have a first version of #30 , we can implement a schema for validating configuration files containing simulation parameters.

jreyns:
Add input validation of sedtrails yml input file.

Added class YAMLConfigValidator which:
- reads an input yml file, and checks it against JSON schema ./resources/config_schema.json
- writes out the schema to a yml to provide users with a complete input example.

Closes #31


resources/config_schema.json
sgpearson17, line 1: I like the way this is broken down into different objects now; it makes sense to me. Looking at this list, I think we have a lot of decisions to make about what to include/not include in this, and how those are structured, but for now I like the format.

resources/sedtrails_config.md
sgpearson17, line 1: so is this the input or the output?

tests/input/test_inputconfigurator.py
sgpearson17, line 1: I ran the test and it passed

sgpearson17 (commented):
Looks good to me! Only thing that I didn't fully understand was whether the provided .md file was the input or the output (or if it's a chicken-and-egg validator/updater kind of thing)

sgpearson17 (commented):
Looks good to me! Only thing that I didn't fully understand was whether the provided .md file was the input or the output (or if it's a chicken-and-egg validator/updater kind of thing)

resources/sedtrails_config.md
jreyns: the markdown table is a modified version of what you pasted in #30, so it is input

resources/create_validation_schema.py
```diff
@@ -0,0 +1,176 @@
...
+    # Optionally, you could add a "required" list at the top level if needed.
+
+    with open(json_schema_filepath, 'w', encoding='utf-8') as f:
+        json.dump(json_schema, f, indent=2)
+
+
+md_file = r"D:\OSS\github_repos\sedtrailsdev\resources\sedtrails_config.md"
+
+# Save to a JSON file
+output_file = r"D:\OSS\github_repos\sedtrailsdev\resources\config_schema.json"
```
jreyns: Note to self: this should be removed and made a relative path in the repo

resources/create_validation_schema.py
```diff
@@ -0,0 +1,176 @@
...
+    type_str = type_str.strip().lower()
+    if type_str == "integer":
+        return "integer"
+    elif type_str == "double":
+        return "number"
+    elif type_str == "string":
+        return "string"
+    elif type_str == "boolean":
+        return "boolean"
+    return "string"
```
manuGil, line 16: Would be using Enums could have been useful in this case? 
About enums: https://docs.python.org/3/library/enum.html 
This is to only for reference, not change is required.

resources/create_validation_schema.py
```diff
@@ -0,0 +1,176 @@
...
+    if required:
+        schema_obj["required"] = required
+    return schema_obj
+
+def md_table_to_json_schema_template(md_filepath, json_schema_filepath):
+    """
+    Reads a Markdown table file and converts it into a JSON Schema file.
+    Chapters (lines starting with [) become groups in the schema and parameter
+    rows are mapped to schema properties.
+    """
```
manuGil, line 67: I think, I missed something here. Should not this function read a YAML file instead? 
The Markdown table was only a place to collect information about the parameters, but I think we should not expect user to write down a table.

src/sedtrails/converters/inputconfiguration.py
```diff
@@ -0,0 +1,206 @@
+import os
+import yaml
+import jsonschema
```
manuGil: `jsonschema` is missing from the TOML file. Could you please add it as follows to the `dependencies` keyword.  e.g. `dependencies = ["jsonschema"]`. 

We should keep the TOML and the conda environment files up to date with no standard dependencies.

src/sedtrails/converters/inputconfiguration.py
```diff
@@ -0,0 +1,206 @@
...
+from typing import Any, Dict, Optional
+
```
manuGil: For the sake of consistent with open source coding styles. I suggest we follow [PEP8](https://peps.python.org/pep-0008/#blank-lines). For example, your code here should have two blank lines before a class. 

A tool like (Flake8)[https://flake8.pycqa.org/en/latest/#] helps to check code against the conversions. There is also [Black](https://pypi.org/project/black/), for automatic formatting, but Flake8 gives you more control.

P.S. If you use Black, do not do automatic formatting for all Python files in the repo. Undesired things may happen (conflicts) ;-)

src/sedtrails/converters/inputconfiguration.py
```diff
@@ -0,0 +1,206 @@
...
+        """
+        # Load YAML file.
+        try:
+            with open(self.yml_filepath, 'r') as f:
+                data: Dict[str, Any] = yaml.safe_load(f)
+        except Exception as e:
+            raise ValueError(f"Error reading YAML file: {e}")
+
+        # Determine and use the appropriate validator for the schema.
+        ValidatorClass = jsonschema.validators.validator_for(self.schema)
```
manuGil: I might be wrong, but I could not find where you declared `self.schema` as an attribute of the `YAMLConfigValidator` class

resources/sedtrails_config.md
```diff
@@ -0,0 +1,231 @@
+| Variable                                   | Type    | Compulsory | Default Value                          | Comment                                                                                                                                                                                              |
```
manuGil, line 1: I do not see the need to use Markdown to define the configurations parameters. I think we agreed to use YAML for this.  That is, the user defines the configuration in a YAML file, but from this file and some of the code I understand that is not the case.  My intention to ask for a Markdown table in #30 was only to gather the information, I wasn't suggesting to use the Markdown itself. Let's clarify this before continuing.

environment.yml
```diff
@@ -12,3 +15,7 @@ dependencies:
   - sphinx=8.2.0
   - myst-parser=4.0.1
   - sphinx-rtd-theme
+  - pyyaml
+  - jsonschema
+  - re
+  - typing
```
manuGil: Aren't these dependencies part of the standard library of Python?

resources/config_schema.json
```diff
@@ -0,0 +1,990 @@
+{
```
manuGil, line 1: I cannot comment on this yet because I am not familiar yet with the way simulations in Sedtrails are defined.

src/sedtrails/converters/inputconfiguration.py
manuGil: `numpy` is also missing in TOML

src/sedtrails/converters/inputconfiguration.py
```diff
@@ -0,0 +1,206 @@
...
+        errors = sorted(validator.iter_errors(data), key=lambda e: e.path)
+        if errors:
+            raise ValueError(f"YAML file validation error: {errors[0].message}")
+
+        # Apply defaults. Pass the full data as root_data for resolving references.
+        self._apply_defaults(self.schema, data, data)
+        self.config = data
+        return self.config
+
+    def export_schema_to_yaml(self, output_file: Optional[str] = None) -> str:
```
manuGil: A test for this should also be included in `test_inputconfigurator.py`. You can test a file has been created and written using the [Pytest temporary directories](https://docs.pytest.org/en/6.2.x/tmpdir.html)

manuGil (changes requested):
I added comments on code styles, and also about missing dependencies and tests. Could you please take a look a those?

I have a concern regarding the code about parsing Markdown tables. Let's discuss this before writing more code related to that.

resources/create_validation_schema.py
jreyns: This is a convenience function I wrote to convert the copy-pasted markdown table that Stuart made with all the parameters he had from the Matlab version Sedtrails into the JSON schema I want to use for the validation. I did not intended this to be part of the 'official' code, that is why I parked it in the resources folder. When we have the structure you made merged in the dev branch, I intend to move the json file to the validator calss folder

src/sedtrails/converters/inputconfiguration.py
aysunrhn: flake8 is listed in the dependencies as well, both the environment file and the project toml. So a +1 for flake8

environment.yml
aysunrhn: re is but I think the others are not, I see only typer is in the `pyproject.toml`. 

https://docs.python.org/3/library/index.html

src/sedtrails/converters/inputconfiguration.py
aysunrhn: Should I open an issue for this and write the test?

src/sedtrails/converters/inputconfiguration.py
jreyns: i will synchronize environment.yml with the [dependencies] in the project toml

src/sedtrails/converters/inputconfiguration.py
jreyns: I ran flake8 on the repo, and I get a lot 'line too long' indications. Do we want to really cut down lines to 79 characters? For formatting, I'd use `ruff` in vs code, but I assume the same holds as for Black regarding merge conflicts?

src/sedtrails/converters/inputconfiguration.py
jreyns: You are totally right. That part dropped off between version I was pulling together, and I also missed it in the tests (because I added fake data there)

resources/sedtrails_config.md
jreyns: This is for our own reference, and to have human readable lists (although I want to dispute 'human-readable' for markdown tables, what a mess). So this will not be used in the code (and hence it is in resources, not sedtrails/src)

environment.yml
jreyns: typing is since 3.5; I will remove them from the env

src/sedtrails/converters/inputconfiguration.py
jreyns: yes please

environment.yml
aysunrhn: Thank you for the heads up!

src/sedtrails/converters/inputconfiguration.py
sgpearson17: Can we perhaps discuss this in the next meeting (briefly, at least)? I feel like if this is a sort of "best practice" that the newbies among us should adopt before we develop too many bad habits, it would be good to clarify the preferred path forward. 😊

resources/create_validation_schema.py
manuGil: @jreyns Alright, thanks for the explanation.

src/sedtrails/converters/inputconfiguration.py
manuGil: I think it will be good to adopt the 79 characters limit for code lines. It will feel very restrictive at the beginning but I think it will better for readability in the long run.

pyproject.toml
```diff
@@ -6,6 +6,17 @@ build-backend = "hatchling.build"
...
 dependencies = [
+   "jsonschema",
+   "numpy",
+   "pyyaml",
+   "pandas",
+   "matplotlib",
+   "xarray",
+   "typer",
+   "sphinx=8.2.0",
+   "myst-parser=4.0.1",
```
manuGil: There is a typo for the versions here. In this file versions are set with `==` @sgpearson17 merged a fix today #91

manuGil (approved):
Thank you for the changes. We  still have the discussion on coding styles, but I will approve this PR. We can discuss further on that later.

pyproject.toml
jreyns: Solved during merge

PR #91, opened 2025-03-05
updated sphinx and myst-parser lines with ==

sgpearson17:
as discussed


manuGil (approved):
This was my mistake. Thanks for the fix

PR #101, opened 2025-03-17
Add pythonpath to pyproject.toml

### LINKED ISSUE #98: Update project.toml to add pytest config for conda
When SedTRAILS environment is created using conda without installing the sedtrails pypi package using pip, pytest failes to import sedtrails packages with the current project directory style.

## Solution
Update the `project.toml` file to explicitly define the python scripts source path

aysunrhn:
Add pythonpath to explciitly define the python scripts source path. This circumvents the need to install a dummy sedtrails package using pip inside a conda environment.

Closes #98


jreyns (approved):
No comments, this works on my system as well in a clean env, so I assume this works everywhere

PR #106, opened 2025-03-19
Progress on R function and one pytest on it

monica-aguil:
The pytest intends to check that Rb, Rs and R contain zeros. They are matrices with the same dimensions.


.vscode/settings.json
manuGil, line 1: This file is about your VS code settings, and it is not relevant for the source code. Could you please remove it from this pull request. You can look at `git revert COMMIT FILENAME` to do that.

src/sedtrails/transport_converter/plugins/physics/library/R_trial.py
```diff
@@ -0,0 +1,74 @@
...
+    theta_cr_A: float,
+    w_s_t: float,
+    u_star_m: NDArray[np.float64],
+    u_star_max: NDArray[np.float64],
+    Uc_mag: NDArray[np.float64],
+    Ub: NDArray[np.float64],
+) -> NDArray:
+    
+    """
+    Computes a reduction factor which is applied on the flow velocities to obtain grain velocities
```
manuGil, line 19: I suggest to keep docstrings with in the context of the code, and avoid additional explanations. For example, "Computes the reduction factor used by grain velocities computations" A way to think on better docstrings is to answer the question "What does this code/class/function do?"

tests/transport_converter/plugins/physics/library/test_R_trial.py
```diff
@@ -0,0 +1,57 @@
+import numpy as np
+from numpy._typing._array_like import NDArray
+import pytest
+from sedtrails.transport_converter.plugins.physics.library.R_trial import Rfactor
+
+# Case that we will be sure that R has to come out as zero
+@pytest.fixture
+def sample_data_case_1():
```
manuGil, line 8: It will be good to know what this data sample represents. It it a data input (which one)? is it a intermediate output, which one? Is it something expected out of a computation or a data file, which ones? This will help other developers to understand better what is being tested.

manuGil (changes requested):
I made several comments about this code. Can you please follow up on them?

manuGil:
@monica-aguil  do you have new changes to this PR. If not, I will merge it as is and we can come back to it latter.

Issue #96, opened 2025-03-12
Internal Data Strucure for SedTRAILS


manuGil:
We need to implement a data class to structure the input data for the simulations. This data structure should have the following attributes.  Data class should hold data for time steps:  n-time, n-1 time and n+1 time.  

Provided by input: (spatial field)
Time: seconds
X: meter
Y: meter
Size time step: time length in second
Bed level: meter
Depth average flow velocity: m/s
Number fractions: 
Bed load sediment transport: kg/m/s (vector)
Suspended sediment transport:  kg/m/s (mass)
Water depth: m
Mean bed shear stress:  pascal
Max bed shear stress: pascal
Suspended sediment concentration: kg/m^3
Non-linear wave velocity: m/s (vector)

aysunrhn:
Interpolated value:
- x
- y
- bed load
- flow velocity
- water level
- concentration
- non-linear velocity
- bed load
- suspended sediment transport
- suspended sediment concentration
- water depth
- mean bed shear stress

PR #110, opened 2025-03-26
Internal Data structures

manuGil:
This PR contains several data classes for the internal data structure for SedTRAILS #96 :

- Implements data structure according to the team's discussion.
- Implements unit test for some scenarios. 
- Particles types are missing details, TODOs were added where this is he case. We can start a discussion about. 
- Additionally, a configuration file for Ruff.

**Could you please help me to define the details for the classes?** For example, from the list of properties in #96  is not clear to me which are *interpolated values* and which are *physical properties*


npannozzo (approved):
From my side I don't see anything missing so I am going to approve. Thanks Manuel!

src/sedtrails/particle_tracer/particle.py
```diff
@@ -0,0 +1,225 @@
...
+
+@dataclass
+class Time:
+    """
+    Class representing a time step in the simulation.
+
+    Attributes
+    ----------
+    time : float
+        The time of the time step in seconds.
```
jreyns: ...since the reference date in the datafile

src/sedtrails/particle_tracer/particle.py
```diff
@@ -0,0 +1,225 @@
...
+    """
+
+    time: float
+
+    def __post_init__(self):
+        if not isinstance(self.time, (int, float)):
```
jreyns: I think we always want floats, not ints

src/sedtrails/particle_tracer/particle.py
```diff
@@ -0,0 +1,225 @@
...
+    Class representing the position of a Particle on a particular time.
+
+    Attributes
+    ----------
+    x : float
+        The x-coordinate of the particle in meters.
+    y : float
+        The y-coordinate of the particle in meters.
+    time : Time
+        The time of the position in seconds.
```
jreyns: same remark, always relative time

src/sedtrails/particle_tracer/particle.py
```diff
@@ -0,0 +1,225 @@
...
+
+    """
+
+    x: float
+    y: float
+    time: Time
+
+
+    def __post_init__(self):
+        if not isinstance(self.x, (int, float)):
```
jreyns: Same remark, I think we should always enforce floats

src/sedtrails/particle_tracer/particle.py
```diff
@@ -0,0 +1,225 @@
...
+    #TODO: define the physical properties of the passive particles
+    physical_properties: dict = field(default_factory=dict)
+
+    def __post_init__(self):
+        # TODO: validate data types once the physical properties are defined
+        pass
+
+
+@dataclass
+class InterpolatedValue:
```
jreyns, line 236: Question here: if we make this an object associated with a Particle object, are we not boxing ourselves in if we want to run operations on np vectors for speed and efficiency? I don't see directly how we can pass this as vectors to functions as arguments

src/sedtrails/particle_tracer/particle.py
```diff
@@ -0,0 +1,225 @@
...
+    """
+    Class for storing interpolated values of a Particle.
+
+    Attributes
+    ----------
+    bed_level : float
+        The bed level of the particle in meters.
+    averaged_velocity : float
+        The averaged flow velocity of the particle in m/s.
+    no_fractions : int
```
jreyns, line 246: this is a constant, and determines what type the particle is. So this is not a property of a Particle object

src/sedtrails/particle_tracer/particle.py
```diff
@@ -0,0 +1,225 @@
...
+        The number of fractions of the particle.
+    bed_load : ndarray
+        The bed load sediment transport of the particle in kg/m/s.
+    sediment_transport : float
```
jreyns: this should read suspended_load

jreyns (changes requested):
Can we go over this together next week?

src/sedtrails/particle_tracer/particle.py
aysunrhn: Make sure to know the reference time to calibrate the time variable. This conversion should be done earlier when converting to the SedTRAILS data structure. 

Add it to the docstring as well.

src/sedtrails/particle_tracer/particle.py
aysunrhn: Should we have an interpolated value class that stores vectors of values instead of floats? This could be computationally more efficient than parallelizing per particle (every particle has a single interpolated value).

Note: we need a way to keep track of which interpolated value corresponds to which particle.

src/sedtrails/particle_tracer/particle.py
aysunrhn: sediment_transport -> suspended_load

src/sedtrails/particle_tracer/particle.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+class Physics:
+    """
+    Class for storing physics converted values for Particles.
+
+    Attributes
+    ----------
+    """
+    # TODO: define which attributes are needed for the Physics class
```
aysunrhn, line 288: How does this class relate to the Fluid and Particle classes that @npannozzo defined for the transport converter?

aysunrhn:
I have made the changes @jreyns requested. But there are a few points from this PR that we should discuss tomorrow.

src/sedtrails/particle_tracer/particle.py
manuGil: Let's discuss this in person. I cannot fully assess the case. But I think I get the main point: accessing the values from an object might create overhead during processing. Is that correct?

src/sedtrails/particle_tracer/particle.py
manuGil: If that's the case, then we should make it an attribute of Particle, and have its value defined by each particle type

src/sedtrails/particle_tracer/particle.py
manuGil: I not sure enforcing floats might be a good idea here and in the cases above. It might create a nuisance for some users because Python internally handles type conversions between integers and floats automatically.  Python will always convert an integer to a float when the case requires it, for example `1 + 1.5 = 2.5`. **Do we want components using the data structure to do type conversions?** if we enforce it here, the it would be the responsibility of those components to do the conversions, when required.

src/sedtrails/particle_tracer/particle.py
manuGil: Output of discussion. The no_fractions, should keep track to the number of fractions (per simulation) per particle type. TODO: workout where this is is best implemented.

manuGil:
This need improvements after today's discussion. We concluded that this data structure isn't correct for SedTrails. Some data classes aren't relevant for the particule-level computations. And others are relevant but at the particle integrations level. As a result, I will modify the data structures. Changes will be made as part of this pull request.

src/sedtrails/particle_tracer/particle.py
manuGil: That's a good point we should discuss this next

manuGil:
I updated the data structure. This closes #131

PR #111, opened 2025-04-02
ST-81: add diffusion library

### LINKED ISSUE #81: Add generic diffusion plugin/function for interpolator
As discussed in #76 , ideally we should break out the random walk diffusion function into a separate function of sorts so that we can have a library of different diffusion methods. e.g. anisotropic diffusion or diffusion that scales with other scalar fields (if available). I'm not saying we should build those now, but if we have a separate function for random walk diffusion that sort of serves as a template, and we have a nice place to slot in [diffusion method] into the interpolator code, it will be easier to make other modifications later on.

jreyns:
Adds a generic method to add diffusion to the particle position vectors. 

Adds a library with two example diffusion classes, implemented using a Strategy pattern:
- Random diffusion as in matlab version Sedtrails
- Diffusion based on gradient of the 2D velocity field

Adds tests for the Diffusion class

closes #81


aysunrhn:
This PR has some conflicts, is it possible if you can look into it @jreyns?

jreyns:
Merge conflicts solved, and pytest is green again

aysunrhn (approved):

PR #115, opened 2025-04-08
Fix most of the pylint errors in tests

aysunrhn:
First batch of pylint error fixes for test modules. There should be no change in the content or the structure.

### DIFF
--- a/tests/test_particle_tracer/test_particleinterpolator.py
+++ b/tests/test_particle_tracer/test_particleinterpolator.py
@@ -1,13 +1,15 @@
+"""
+Unit tests for the ParticlePositionCalculator class, covering field interpolation, 
+particle updates with and without diffusion, parallel vs serial consistency, 
+geographic adjustments, and mixed network interpolation.
+"""
 import numpy as np
 import pytest
-
 from sedtrails.particle_tracer.position_calculator import ParticlePositionCalculator
 
 # -----------------------------------------------------------------------------
 # Fixtures
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


tests/test_particle_tracer/test_particleinterpolator.py
```diff
@@ -93,7 +98,7 @@ def test_update_particles_with_diffusion(simple_interpolator):
     dt = 0.1
     rndfac = 0.5
     np.random.seed(42)
-    x_new, y_new, xdiff, ydiff = simple_interpolator.update_particles(
+    _, _, xdiff, ydiff = simple_interpolator.update_particles(
```
jreyns: this test can be deleted, as this is now covered in #81

jreyns (changes requested):
we need to remove the diffusion test. we can do it here, or make a new PR

tests/test_particle_tracer/test_particleinterpolator.py
aysunrhn: This is the part that is causing the conflict, right? So I think we can remove it here then I will merge the PR.

jreyns: yes, please do

aysunrhn: I resolved it now, please check the file again to make sure it is OK.

jreyns (approved):
This should be okay

manuGil:
I will merge this then.

PR #123, opened 2025-04-16
Fix ruff errors except for the exceptions

### LINKED ISSUE #122: Fix ruff errors in existing code
Now we are using ruff for linting, the existing code needs to be checked and fixed to comply with the ruff rules we set.

All errors excluding those relating to the exceptions.

aysunrhn:
First attempt to fix linting errors in existing code in `src/sedtrails/sedtrails.py` and `tests/test_transport_converter/test_format_converter.py`. I will wait until we define custom exceptions for SedTRAILS before I fix the linting errors related to those. Closes #122 

Files modified:
- `src/sedtrails/sedtrails.py`
- `tests/test_transport_converter/test_format_converter.py`

@jreyns Please have a look at the function `network_analysis`: the input and output file paths were the same in the original script but based on the help message I changed the output file path. 
https://github.com/aysunrhn/sedtrails/blob/bbc44bbca302f53531f5cd8d482451fd6566578f/src/sedtrails/sedtrails.py#L145-L169

### DIFF
--- a/src/sedtrails/sedtrails.py
+++ b/src/sedtrails/sedtrails.py
@@ -6,17 +6,13 @@ app = typer.Typer(
     help="Sedtrails: Configure, run, and analyze sediment particle tracking."
 )
 
+CONFIG_FILE = "sedtrails.yml"
+PARTICLE_TRACK_FILE = "sedtrails.nc"
+ANALYSIS_FILE = "analysis.nc"
 
 # Subcommand to load and validate a YAML configuration file.
 @app.command("load-config")
-def load_config(
-    config_file: Path = typer.Option(
-        "sedtrails.yml",
-        "--config",
-        "-c",
-        help="Path to the SedTRAILS configuration file.",
-    ),
-) -> dict:
+def load_config(config_file: Path = None) -> dict:
     """
     Load and validate a YAML configuration file using an existing validator.
     Returns a dictionary with the valid configuration settings.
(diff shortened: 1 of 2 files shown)


manuGil:
@aysunrhn I added some example of custom exceptions. However, I did not define exception for this particular code. Is that something you think it is necessary?

aysunrhn:
> @aysunrhn I added some example of custom exceptions. However, I did not define exception for this particular code. Is that something you think it is necessary?

It might be useful to define custom exceptions for errors relating to the configuration file (error validating/loading configuration file). I'm not sure if it makes sense to define exceptions for failed simulation runs, maybe there are specific types of failures that we can inform the user about their failed simulation runs. I suggest we add this to the agenda next week to discuss.

manuGil:
> > @aysunrhn I added some example of custom exceptions. However, I did not define exception for this particular code. Is that something you think it is necessary?
> 
> It might be useful to define custom exceptions for errors relating to the configuration file (error validating/loading configuration file). I'm not sure if it makes sense to define exceptions for failed simulation runs, maybe there are specific types of failures that we can inform the user about their failed simulation runs. I suggest we add this to the agenda next week to discuss.

I implemented a few custom exceptions in the exceptions module. I need to merge this to avoid conflicts with #126 Please continue the discussion about the `network_analysis` in the new issue #127

manuGil (approved):
This can be merge as other PR depends on it.

PR #126, opened 2025-04-16
Exception Examples

manuGil:
This RP add examples of how to define and use custom exceptions in Python. 

- As we develop the code we should be attentive to when we should prioritize the use of custom exceptions over built-in exception, with the purpose of providing better error messages to the end user.
- It  contains a few custom exceptions for the CLI and YAML parset.
- It allow mutable function calls when using `typer.Option()`


manuGil:
I am merging this to avoid generating conflicts.

PR #142, opened 2025-04-30
Nefis Parser

manuGil:
This PR contains work in progress for a parser for binary files from the Delft3D software.

## Progress so far

- Implement `vs_use` and `vslet` scripts in a similar way as the MatLab implementation. 
- These scripts need to be tested and debuged.
- Fix to the _open_data_file method in the `vs_use.py`.

## Current issue
The following error is raised when calling `_read_group_data_()` in `vs_use`. The code fails to decode some binary characters using ASCII. This could be a consequence of not parsing the file properly in previous steps (e.g. for example if the type and endianness is not being identified correctly. Or the hash table used to identified the data groups aren't been calculated correctly. 

```shell
Exception has occurred: UnicodeDecodeError
'ascii' codec can't decode byte 0xbc in position 3: ordinal not in range(128)
  File "/Users/mgarciaalvarez/devel/sedtrails/src/sedtrails/nefis/vs_use.py", line 204, in _read_group_data
    ianames = [f.read(16).decode('ascii').strip('\x00') for _ in range(5)]
               ~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/Users/mgarciaalvarez/devel/sedtrails/src/sedtrails/nefis/vs_use.py", line 84, in __init__
    self._read_group_data()
    ~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/mgarciaalvarez/devel/sedtrails/src/sedtrails/nefis/vs_use.py", line 283, in <module>
    s = VSUse(r'/Users/mgarciaalvarez/devel/sedtrails/sample-data/trim-f34.dat')
UnicodeDecodeError: 'ascii' codec can't decode byte 0xbc in position 3: ordinal not in range(128)
```

I use the files in the NEFIS Python repository  (`trim-f34.dat` and trim-f34.def) for this. For which the following type and endianness was determined by the code in this pull request: 
- AddressType: **uint32**
- Format: **big endian**

## Next steps
- I want to check if AddressType and Format for the data file are correct. I need a way to check the format and type of the file indenpendently, 
- I want to check if for that file an ASCII encoder is the correct choice. I tried UTF-8 and Latin1, but they produced gibberish. Again I need a way to check the encoding of the files to be  sure. 
- Another part where I have doubts is in the generation of the hash table in `_read_group_data()` I do not understand the many constants used there, but  my guess is that understanding that is key to know what might be wrong.


manuGil:
Note that the files from Nefis-Python have header that mentions version 4.0 of Delf3D. But the files provided by @sgpearson17 have a header that mention version 5.0. fixing the parser for one version might not work for the other.

jreyns:
I think we can safely assume that people use 5.0

manuGil:
> I think we can safely assume that people use 5.0

Then, I will focus debugging the code for this version

manuGil:
We will come back to this later and move ahead with DFM.

sgpearson17:
Closed as per discussion on 2026-04-29 with Johan. We will implement offline matlab d3d4 to nc converter and make that a new input type.

PR #148, opened 2025-05-07
Scaffolding Configuration Interface

manuGil:
This is an interface for the component in the Configuration interface


jreyns (approved):
Approved as basis for project scaffolding

PR #149, opened 2025-05-07
ST-147: configuration controller class

### LINKED ISSUE #147: Discussion about configuration interface component.
Discuss the current state and possible refactoring to the code in  this component with @jreyns.

jreyns:
Add implementation of configuration controller class 
Closes #147

### DIFF
--- a/src/sedtrails/configuration_interface/configuration_controller.py
+++ b/src/sedtrails/configuration_interface/configuration_controller.py
@@ -6,7 +6,10 @@ Reads simulation configuration files, applies default configuration values,
 and provides configurations to other components.
 """
 
+import os
 from abc import ABC, abstractmethod
+
+from sedtrails.configuration_interface.validator import YAMLConfigValidator
 
 
 class Controller(ABC):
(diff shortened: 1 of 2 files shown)


src/sedtrails/configuration_interface/configuration_controller.py
```diff
@@ -16,10 +19,9 @@ class Controller(ABC):
 
     # TODO: Something to keep in mind in this class is that
     # if we can and it is convenient to have separted methods to retrieve parts of the configuration
-    # that go to the particle tracer and the transport converter. For example, if it is convenient to 
+    # that go to the particle tracer and the transport converter. For example, if it is convenient to
     # have a method that returns the configuration values that are only relevant for the particle tracer.
```
manuGil, line 23: we can remove this comments. We agreed to pass all configurations to other components.

src/sedtrails/configuration_interface/configuration_controller.py
```diff
@@ -57,11 +55,16 @@ def read_config(self, config_file: str) -> None:
             The path to the configuration file to read.
         """
 
-        #TODO: Implement reading from a configuration file
-        # read the configuration file
-        # validate the configuration
+        valid_json = os.path.join(r'/config_schema.json')  # not happy with hardcoded path
```
manuGil, line 58: The way I understand it is that the `config_schema.json` is part part of the source code; therefore, there is no problem with hard-coding the path.

manuGil (approved):
@jreyns This is looks good, I only made a comment. I fixed the Ruff error in a pull request I will create later today, so this can be merge as is.

manuGil (approved):
@jreyns This is looks good, I only made a comment. I fixed the Ruff error in a pull request I will create later today, so this can be merge as is.

manuGil (approved):
@jreyns This is looks good, I only made a comment. I fixed the Ruff error in a pull request I will create later today, so this can be merge as is.

PR #156, opened 2025-05-14
documentation (old and benchmark runs)

sgpearson17:
existing documentation (old user manual and paper draft) plus new information about matlab benchmark runs


aysunrhn (commented):
I noticed there are 2 empty files:

- `docs/user/jan2023/old_overview.md`
- `docs/user/nov2023/old_sedtrailsPaperDraft.md` 

Are these meant to be placeholders? Or did something go wrong when uploading them?

sgpearson17:
Indeed something seems to have gone wrong while uploading. I was copy-pasting a bunch of stuff and must have missed something. I will try to fix in the coming days

Issue #151, opened 2025-05-07
Implement NetCDF Reader with xugrid Integration to create transport field data structure


bartvanwesten:
### Implement NetCDF Reader with xugrid Integration to create transport field data structure

**Objective**: Develop a module to read NetCDF files adhering to UGRID conventions and convert them into `xugrid.UgridDataset` objects.([[deltares.github.io](https://deltares.github.io/xugrid)][1])

**Details**:

* **Module Location**: `src/sedtrails/transport_converter/format_converter.py`
* **Functionality**:

  * Utilize `xugrid.open_dataset()` to load NetCDF files.
  * Ensure compatibility with UGRID-compliant NetCDF files.

**References**:

* xugrid Documentation: [https://deltares.github.io/xugrid/examples/quick\_overview.html\:contentReference\[oaicite:23\]{index=23}]([https://deltares.github.io/xugrid/examples/quick_overview.html:contentReference[oaicite:23]{index=23}](https://deltares.github.io/xugrid/examples/quick_overview.html:contentReference%5Boaicite:23%5D%7Bindex=23%7D))

**Next Steps**:

* [ ] Implement the NetCDF reading functionality.
* [ ] Integrate the reader into the specified module.
* [ ] Build a generic data structure for transport and flow fields

manuGil:
I like the way this issue is documented (the structure). Can you make this a template for all future issues?

PR #157, opened 2025-05-14
Implement SedTRAILS format_converter module for NetCDF processing

bartvanwesten:
Created a new format_converter module for the SedTRAILS particle tracking system  #151 

- Designed SedtrailsData structure for multidimensional hydrodynamic data with efficient time handling
- Implemented FormatConverter class to read and transform Delft3D Flexible Mesh NetCDF files
- Added robust variable mapping from DFM to SedtrailsData format
- Built time-dimension handling with time as first axis in all time-dependent variables
- Included dictionary-based storage for vector quantities (velocity, sediment transport)
- Added time indexing functionality for easy access to specific time slices
- Implemented comprehensive error handling for missing files and variables
- Added fallback to xarray when xugrid can't process a file
- Created command-line interface for testing with sample NetCDF files
- Added detailed docstrings and comments throughout the code

This module serves as the foundation for reading hydrodynamic simulation outputs and preparing them for particle tracing, supporting the overall SedTRAILS coastal sediment modeling workflow.

### DIFF
--- a/src/sedtrails/particle_tracer/data_retriever.py
+++ b/src/sedtrails/particle_tracer/data_retriever.py
@@ -4,10 +4,9 @@ Flow Field Data Retriever
 Retrieves flow field data from the transport converter or the simulation caching and state tracker.
 Performs temporal interpolation to provide accurate flow field data at any time point for particle tracing.
 """
-from typing import Dict, Tuple, Optional, Callable
+from typing import Tuple
 import numpy as np
 import sys
-import os
 from pathlib import Path
 
 # Add parent directory to path to allow importing from other packages
--- a/src/sedtrails/particle_tracer/data_retriever.py
+++ b/src/sedtrails/particle_tracer/data_retriever.py
@@ -1,5 +1,165 @@
 """
 Flow Field Data Retriever
-=========================
-Retrieve flow field data from the transport converter or the simulation caching and state tracker.
+
+Retrieves flow field data from the transport converter or the simulation caching and state tracker.
+Performs temporal interpolation to provide accurate flow field data at any time point for particle tracing.
 """
+from typing import Dict, Tuple, Optional, Callable
+import numpy as np
+import sys
+import os
+from pathlib import Path
+
+# Add parent directory to path to allow importing from other packages
+sys.path.append(str(Path(__file__).parent.parent.parent))
+from sedtrails.transport_converter.format_converter import SedtrailsData
+
+
+class FlowFieldDataRetriever:
+    """
+    Retrieves and interpolates flow field data from SedtrailsData.
+    
+    This class provides methods to extract flow fields at specific times,
+    performing temporal interpolation as needed. Currently supports depth-averaged
+    flow velocity.
+    """
+    
+    def __init__(self, sedtrails_data: SedtrailsData):
+        """
+        Initialize the FlowFieldDataRetriever.
+        
+        Parameters:
+        -----------
+        sedtrails_data : SedtrailsData
+            The SedtrailsData object containing the flow fields
+        """
+        self.sedtrails_data = sedtrails_data
+        
+        # Only using depth-averaged flow velocity for now
+        self.flow_field_name = "depth_avg_flow_velocity"
+    
+    def get_interpolation_indices(self, target_time: float) -> Tuple[int, int, float]:
+        """
+        Get indices for interpolation between two time steps.
+        
+        Parameters:
+        -----------
+        target_time : float
+            Target time in seconds since reference_date
+            
+        Returns:
+        --------
+        Tuple[int, int, float]
+            (lower_index, upper_index, weight) where weight is the interpolation factor [0-1]
+        """
+        times = self.sedtrails_data.times
+        
+        # Handle edge cases
+        if target_time <= times[0]:
+            return 0, 0, 0.0
+        
+        if target_time >= times[-1]:
+            last_idx = len(times) - 1
+            return last_idx, last_idx, 1.0
+        
+        # Find the index of the last time that is less than or equal to the target time
+        lower_idx = np.searchsorted(times, target_time, side='right') - 1
+        upper_idx = lower_idx + 1
+        
+        # Calculate the interpolation weight
+        time_range = times[upper_idx] - times[lower_idx]
+        
+        # Avoid division by zero
+        if time_range == 0:
+            weight = 0.0
+        else:
+            weight = (target_time - times[lower_idx]) / time_range
+            
+        return lower_idx, upper_idx, weight
+    
+    def get_flow_field(self, time: float) -> Tuple[np.ndarray, np.ndarray]:
+        """
+        Get the flow field at the specified time.
+        
+        This method performs temporal interpolation between the two closest time steps
+        in the SedtrailsData object to obtain the flow field at the requested time.
+        
+        Parameters:
+        -----------
+        time : float
+            Time in seconds since the reference date of the SedtrailsData object
+            
+        Returns:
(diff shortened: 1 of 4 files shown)


src/sedtrails/particle_tracer/data_retriever.py
```diff
@@ -1,5 +1,164 @@
...
+        """
+        times = self.sedtrails_data.times
+        
+        # Handle edge cases
+        if target_time <= times[0]:
+            return 0, 0, 0.0
+        
+        if target_time >= times[-1]:
+            last_idx = len(times) - 1
+            return last_idx, last_idx, 1.0
```
manuGil: The 1.0 here is a 'magic number'. Magic numbers are constants that have no context (we do know know what they represent,  or why is not 2.0 instead of 1.0), and their use should be avoided. A way to avoid them is to use symbolic constants. See the Python example here: https://refactoring.guru/replace-magic-number-with-symbolic-constant

src/sedtrails/particle_tracer/data_retriever.py
```diff
@@ -1,5 +1,164 @@
...
+        
+        # Find the index of the last time that is less than or equal to the target time
+        lower_idx = np.searchsorted(times, target_time, side='right') - 1
```
manuGil: Other cases of magic numbers

src/sedtrails/particle_tracer/data_retriever.py
```diff
@@ -1,5 +1,164 @@
...
+        # Get indices for interpolation
+        lower_idx, upper_idx, weight = self.get_interpolation_indices(time)
+        
+        # If time is exactly at a time step or outside the range, no interpolation needed
+        if lower_idx == upper_idx:
+            time_slice = self.sedtrails_data[lower_idx]
+            flow_field = time_slice[self.flow_field_name]
+            return flow_field['x'], flow_field['y']
+        
+        # Otherwise, perform linear interpolation between the two time steps
```
manuGil: This will work, but use an `else` here to ensure the proper execution of the program. It will make debugging easier in the long run.

src/sedtrails/particle_tracer/data_retriever.py
```diff
@@ -1,5 +1,164 @@
...
+        lower_slice = self.sedtrails_data[lower_idx]
+        upper_slice = self.sedtrails_data[upper_idx]
+        
+        lower_flow = lower_slice[self.flow_field_name]
+        upper_flow = upper_slice[self.flow_field_name]
+        
+        # Linear interpolation: result = (1-w)*lower + w*upper
+        flow_x = (1 - weight) * lower_flow['x'] + weight * upper_flow['x']
+        flow_y = (1 - weight) * lower_flow['y'] + weight * upper_flow['y']
+
```
manuGil: To make this code more modular and testable, I suggest to move the interpolation to its own independent function. If doing it here is not feasible. Could you please open an issue so that we can refactor the code later?

src/sedtrails/particle_tracer/data_retriever.py
```diff
@@ -1,5 +1,164 @@
...
+"""
+from typing import Tuple
+import numpy as np
+import sys
+from pathlib import Path
+
+# Add parent directory to path to allow importing from other packages
+sys.path.append(str(Path(__file__).parent.parent.parent))
+from sedtrails.transport_converter.format_converter import SedtrailsData
+
```
manuGil: Although, add the path in this way enables importing the modules in sedtrails. This method may create problems when packaging the code to be installable via PyPI or Conda. 
A fool-proof way is to import sedtrails as we are importing any other package:
`from sedtrails.transport_converter.format_converter .....`

For that to work, you have to install sedtrails in editable mode: 
```shell
# On the root of the repository:
pip install -e .[dev]
```
Can you install the package in editable mode in your set up, and remove lines 12 - 13?

src/sedtrails/transport_converter/format_converter.py
```diff
@@ -1,80 +1,544 @@
...
+    fractions: int
+    bed_load_sediment: Dict[str, np.ndarray]
+    suspended_sediment: Dict[str, np.ndarray]
+    water_depth: np.ndarray
+    mean_bed_shear_stress: np.ndarray
+    max_bed_shear_stress: np.ndarray
+    sediment_concentration: np.ndarray
+    nonlinear_wave_velocity: Dict[str, np.ndarray]
+    
+    def __getitem__(self, time_idx: int) -> Dict:
```
manuGil: this might sound very picky, but  'idx' is only 2 letter away of being more meaningful. Wouldn't be better to use the full word?

src/sedtrails/transport_converter/format_converter.py
```diff
@@ -1,80 +1,544 @@
...
+        Dict
+            Dictionary containing all data for the specified time index
         """
-        return self
-
+        if time_idx < 0 or time_idx >= len(self.times):
+            raise IndexError(f"Time index {time_idx} out of bounds (0-{len(self.times)-1})")
+        
+        # Create a dictionary with time-specific data
+        time_data = {
```
manuGil: time_data sounds misleading, because the dictionary does not contains data about time. I think. Is this the data for a particular time in the dataset? Then, it wold be better to use a generic 'data' as name. The docstring above, makes clear that is only part of the dataset.

src/sedtrails/transport_converter/format_converter.py
```diff
@@ -1,80 +1,544 @@
...
+        # Calculate the interpolation weight
+        time_range = self.times[upper_idx] - self.times[lower_idx]
+        
+        # Avoid division by zero
+        if time_range == 0:
+            weight = 0.0
+        else:
+            weight = (target_time - self.times[lower_idx]) / time_range
+            
+        return lower_idx, upper_idx, weight
```
manuGil: Isn't this code repeated in  `data_retriever.py`? If we need it in both, that's OK. We just have to make sure that the code is not executed twice for the same operation.

src/sedtrails/transport_converter/format_converter.py
```diff
@@ -1,80 +1,544 @@
...
+    
+    def get_interpolation_indices(self, target_time: float) -> Tuple[int, int, float]:
+        """
+        Get indices for interpolation between two time steps.
+        
+        Parameters:
+        -----------
+        target_time : float
+            Target time in seconds since reference_date
+
```
manuGil: I have de feeling that this function should belong somewhere else, because it is part of the interpolation.

src/sedtrails/transport_converter/format_converter.py
```diff
@@ -1,80 +1,544 @@
...
+    print(f"  Number of time steps: {time_info['num_times']}")
+    
+    # Convert all time steps to SedtrailsData
+    sedtrails_data = converter.convert_to_sedtrails_data()
+    
+    # Print some information about the converted data
+    print("\nConverted data:")
+    print(f"  Number of time steps: {len(sedtrails_data)}")
+    print(f"  Reference date: {sedtrails_data.reference_date}")
+    print(f"  Time values range: {sedtrails_data.times[0]} to {sedtrails_data.times[-1]} seconds since reference")
```
manuGil: Can you move these examples to the examples directory? you can create a new python script which import the class and then add the example code. That way, we won't forget that we have an example, and we will document it later on

manuGil (changes requested):
Hello Bart, Thanks for all the work here. I have several comments, some more relevant then others. Please, make changes when relevant.

manuGil (approved):
We can merge this now, and test the example with the other components.

PR #158, opened 2025-05-14
experiments with grain property classes and functions

sgpearson17:
UPDATE: I screwed up my dev branch synchronization so this apparently never got submitted correctly, but here are some of the tests I tried developing for Soulsby. Monica/Natascia, maybe you kept going with this, I'm not sure. Hopefully there are no merge conflicts. In either case, I think this will be shelved for now until we get the "minimum viable product" passive-particle tracing pipeline in place (our July 1st goal). Once we go back to implementing Soulsby we can revisit this. 

ORIGINAL TEXT: I made a first attempt at breaking out the basic grain size property functions used in soulsby (Dstar, shields_cr, shields_number). I have not had time to finish them all yet, and also still want to finish working on the unit tests (using example data from Soulsby (1997) textbook) but am uploading what I have for now so that Monica and Natascia can take a look

### DIFF
--- a/src/sedtrails/transport_converter/plugins/physics/library/dstar.py
+++ b/src/sedtrails/transport_converter/plugins/physics/library/dstar.py
@@ -1,0 +1,28 @@
+import numpy as np
+
+class DStar:
+    def __init__(self, g, rhoS, rhoW, visc_kin, d):
+        """
+        Initialize the DStar class with the given parameters.
+
+        Parameters:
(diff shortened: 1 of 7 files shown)


manuGil:
Shall we keep this a draft? or merge it?

manuGil:
Let's merge this for now, and document in a new issue how software should be written (apply TDD when coming back to this task).

PR #161, opened 2025-05-14
Modify Time class in Particle class

### LINKED ISSUE #160: Update Time class in the SedTRAILS particle data structure
Modify the Time class inside the SedTRAILS Partcile class following our discussions on the internal data structure

## Relevant files
- `src/sedtrails/particle_tracer/particle.py`

aysunrhn:
I modified the Time class in the SedTRAILS internal Particle class following our discussions about how to incorporate time internally. Closes #160 

## Important points to note:
- We will stick to the `numpy.datetime64` in seconds to represent time throughout the simulation
- The time difference (or time offset, or simulation run time) is a `numpy.timedelta64` object in seconds
- Also added preliminary tests for the modified Time class in `test/particle_tracer/test_particle.py` as a new `TestTime` class

## Relevant files:
- `src/sedtrails/particle_tracer/particle.py`
- `src/sedtrails/tests/particle_tracer/test_particle.py`

### DIFF
--- a/src/sedtrails/particle_tracer/particle.py
+++ b/src/sedtrails/particle_tracer/particle.py
@@ -16,8 +16,8 @@ class Time:
 
     Attributes
     ----------
-    reference_date : numpy.datetime64
-        The reference date for calculating time.
+    reference_date : str
+        The reference date as string in format 'YYYY-MM-DD'
     offset_seconds : numpy.timedelta64
         The offset in seconds from the reference date.
 
@@ -29,14 +29,21 @@ class Time:
         Updates the current time by adding a delta in seconds.
     """
 
-    reference_date: np.datetime64 = field(default=np.datetime64('1970-01-01T00:00:00', 's'))
+    # refence_date should be a str, the class will do the transformation into a numpy.datetime64 object
+    reference_date: str = field(default='1970-01-01')
     offset_seconds: np.timedelta64 = field(default=np.timedelta64(0, 's'))
+    _reference_date_np: np.datetime64 = field(init=False)
 
+    def _convert_to_datetime64(self, date_str: str) -> np.datetime64:
+        """Convert reference_date in str format to numpy.datetime64"""
+        return np.datetime64(f"{date_str}T00:00:00", 's')
+    
     def __post_init__(self):
-        if not isinstance(self.reference_date, np.datetime64):
-            raise TypeError(f"Expected 'reference_date' to be a numpy.datetime64, got {type(self.reference_date).__name__}")
-        if self.reference_date.dtype != 'datetime64[s]':
-            raise ValueError("reference_date must have a resolution of seconds (datetime64[s]).")
+        # Accept reference_date in str format TODO: should we allow for non str formats?
+        try:
+            self._reference_date_np = self._convert_to_datetime64(self.reference_date)
+        except ValueError:
+            raise ValueError("reference_date must be in format 'YYYY-MM-DD'")
         if not isinstance(self.offset_seconds, np.timedelta64):
             raise TypeError(f"Expected 'offset_seconds' to be a numpy.timedelta64, got {type(self.offset_seconds).__name__}")
         if self.offset_seconds.dtype != 'timedelta64[s]':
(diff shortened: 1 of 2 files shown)


src/sedtrails/particle_tracer/particle.py
```diff
@@ -16,52 +16,57 @@ class Time:
 
     Attributes
     ----------
-    time_step : int
-        The current time step in a simulation. A non-negative integer.
+    reference_date : numpy.datetime64
```
manuGil: Could you modify the datatype of `reference_date` so that we pass a string with the date, instead of the numpy object? For example: `reference_date= "2025-05-14"`

manuGil: Then, the class does the transformation that are needed.

src/sedtrails/particle_tracer/particle.py
```diff
@@ -16,52 +16,57 @@ class Time:
...
+        The reference date for calculating time.
+    offset_seconds : numpy.timedelta64
```
manuGil: Would be more intuitive to pass pass only the number of units to `offset_seconds` and hard code the units (seconds). That way it is easier for humans to interpret. I am thinking that in the configuration file, something like the size of the time step will be defined in seconds, for example. We could also accept both data types, might be a case in which the offset can be process faster if in deltatime

src/sedtrails/particle_tracer/particle.py
```diff
@@ -16,52 +16,57 @@ class Time:
...
+        -------
+        numpy.datetime64
+            The current time in the simulation.
         """
-        if not isinstance(time_step, int):
-            raise TypeError(f"Expected 'time' to be a float, got {type(time_step).__name__}")
-        if time_step < 0:
-            raise ValueError(f"Expected 'time' to be a non-negative integer, got {time_step}")
-        self.time_step = time_step
+        return self.reference_date + self.offset_seconds
```
manuGil: Can you double check if this the the correct way to calculate the current simulation time? 
According to my conversion with @bartvanwesten, what they need is a way in which the point from which time is calculated can be changed. For example, having the capability to set at which Gregorian date/time the simulation time starts. That is they need to say simulation time starts at 1 May 2025 midnight, therefore, the current time should be the number of second since that date/time. I have the impression that the current code does not enable that.

manuGil: Can you also add a convenient function to get the time in seconds since the reference date/time

tests/particle_tracer/test_particle.py
```diff
@@ -97,3 +98,52 @@ class TestPhysics:
...
+        time_instance = Time(reference_date=reference_date, offset_seconds=offset_seconds)
+
+        delta_seconds = np.timedelta64(120, 's')
+        time_instance.update(delta_seconds)
+
+        expected_time = reference_date + delta_seconds
+        assert time_instance.get_current_time() == expected_time, \
+            f"Updated current time should be {expected_time}, but got {time_instance.get_current_time()}."
+
+    def test_update_subtract_seconds(self):
```
manuGil, line 131: simulation time will move very likely only forwards. But it is good check it can be use either way.

manuGil (changes requested):
@aysunrhn Could you please consider my comments and make changes to the code?

src/sedtrails/particle_tracer/particle.py
aysunrhn: In this case, do you recommend that we make the `offset_seconds` variable either an integer or deltatime object?

src/sedtrails/particle_tracer/particle.py
aysunrhn: If I understand correctly, the reference_date is different from the time the simulation starts. I think I confused the two concepts here.

src/sedtrails/particle_tracer/particle.py
manuGil: Let's make it an integer.  The class should  be responsible of converting the type is needed

src/sedtrails/particle_tracer/particle.py
manuGil: That is correct.  The reference-date defines what is time zero on a simulation run. Then a simulation can start at any time after that. 

For example, we want to be able to say,  for simulation S, the reference_date Rd is 16 May, therefore, the simulation time Ts is 0 at RD, and it progresses  from there.  Then, S can have a starting time St that is N number of second  from RD. 

Does this makes things clearer? If not let's have a chat  next week.

aysunrhn: Thank you for looking into the code!

I changed the code to accept `offset_seconds` as int. Now we have a method to convert it to np.deltatime64 (we should check later if this is computationally more efficient way to go).

I also modified the relevant test functions in `tests/particle_tracer/test_particle.py` to be consistent with the new variable convention. 

It will be good to talk tomorrow to clarify some of the details, to make sure I got it correct.

manuGil (approved):
@aysunrhn  Thank you for adapting the code. I will merge this for now. For further changes we can make a new pull request.

PR #162, opened 2025-05-14
Seeding Tool

manuGil:
This RP creates includes:

- An interface for the Particle seeding tool
- Seeders for the following cases: set initial positions for given x, y positions, and random positions.
- Unit test for XYSeeding, and RandomSeeding


examples/sedtrails_config.yaml
```diff
@@ -0,0 +1,198 @@
+General:
```
sgpearson17, line 1: Great to see this now in actual yaml form. I have a few suggestions for things to change but think that is a matter for future issues and will also depend on what we actually use in the end.

manuGil:
I would like to merge this ASAP to build on Bart's examples in #173  @sgpearson17  or @bartvanwesten  would you be able to make a quick review on this today?

sgpearson17:
yes, going through it as we speak. will be finished very soon

src/sedtrails/particle_tracer/particle_seeding.py
```diff
@@ -1,5 +1,275 @@
...
+        Returns
+        -------
+        list[Particle]
+            List of seeded particles.
+        """
+        pass
+
+    def _validate_initial_positions(self, initial_positions: List[tuple] | tuple, count: int) -> bool:
+        """
+        Validate the initial positions and count of particles.
```
sgpearson17, line 51: I think in future we should add a validation check to ensure that the initial positions all lie within the model domain (i.e. to make sure we are placing particles somewhere where they will actually move)

src/sedtrails/particle_tracer/particle_seeding.py
```diff
@@ -1,5 +1,275 @@
...
+                f'Number of initial positions ({len(initial_positions)}) does not match the count ({count}).'
+            )
+        if isinstance(initial_positions, tuple) and count != 1:
+            validity = False
+            raise ValueError(f'Number of initial positions ({initial_positions}) does not match the count ({count}).')
+        return validity
+
+    def _validate_release_times(self, release_times: list | int, count: int):
+        """
+        Validate the release times and count of particles.
```
sgpearson17, line 95: I think we should also add a check to see whether the release time is within the model simulation time. In early versions of matlab sedtrails, starting a particle too early (before simulation started) would trigger later problems in the model until I built a check like this in. Not sure if this is the place or time for it, but we should do something about that eventually

sgpearson17 (approved):
this looks sensible to me, I think it's ok to merge for now- then we can start to test things with what Bart did

examples/sedtrails_config.yaml
manuGil: Whatever ideas you have, would be useful to write them down as issue. So that we do not forget, or at least we can discuss them.

src/sedtrails/particle_tracer/particle_seeding.py
manuGil: I added this as TODO to the file

src/sedtrails/particle_tracer/particle_seeding.py
manuGil: I think it is best to do these kind of validation in the Configuration Interface (very early in the simulation), in that way obvious input errors can be caught before attempting running a simulation. As a consequence, a user does not have to wait until an invalid value is check this deep in the code. I created #175

sgpearson17: Perfect, sounds great Manuel! thanks for creating the issue.

PR #172, opened 2025-05-21
Update Time class

### LINKED ISSUE #164: CF Compliance of Time Conventions
As per discussion 2025-05-21, we need to come to an agreement on the time convention used by particle.py. Currently a format of yyyy-mm-ddT00:00:00 is used, and after debating the use of JR (Johan Reyns) convention, we decided to stick with CF conventions as per:
https://cfconventions.org/cf-conventions/cf-conventions.html#time-coordinate

aysunrhn:
This PR includes:

- Updates to [`src/sedtrails/particle_tracer/time.py`](https://github.com/sedtrails/sedtrails/blob/dev/src/sedtrails/particle_tracer/particle.py) to follow the CF conventions and `get_seconds_since_reference` to get seconds since reference in a human readable format
- Modifications to [`tests/particle_tracer/test_time.py`](https://github.com/sedtrails/sedtrails/blob/dev/tests/particle_tracer/test_particle.py) to add unit test for the `get_seconds_since_reference` method

Closes #164

### DIFF
--- a/src/sedtrails/particle_tracer/time.py
+++ b/src/sedtrails/particle_tracer/time.py
@@ -65,12 +65,22 @@ class Time:
         total_seconds = self.start_time + delta_seconds
         return self._reference_date_np + np.timedelta64(total_seconds, 's')
 
-    def get_seconds_since_reference(self) -> str:
+    def get_seconds_since_reference(self, delta_seconds: int = 0) -> int:
         """
-        Returns the number of seconds since the reference date/time in a 
-        human readable format
+        Returns the current time in simulation as integer in units of seconds.
+
+        Parameters
+        ----------
+        delta_seconds : int, optional
+            Additional seconds to add to the current simulation time (default is 0).
+
+        Returns
+        -------
+        total_seconds : int
+            The total seconds in the simulation since the reference date.
         """
-        return str(self.get_current_time())   
+        total_seconds = self.start_time + delta_seconds
+        return total_seconds
 
     def time_as_timedelta64(self) -> np.timedelta64:
         """
--- a/src/sedtrails/particle_tracer/time.py
+++ b/src/sedtrails/particle_tracer/time.py
@@ -4,7 +4,8 @@ Time class used internally to represent simulation time.
 
 from dataclasses import dataclass, field
 import numpy as np
-
+import re
+from sedtrails.exceptions.exceptions import ReferenceDateFormatError
 
 @dataclass
 class Time:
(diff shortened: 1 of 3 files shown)


aysunrhn:
See my comment in issue #164, and confirm that this addresses the issue.

> In numpy, `YYYY-MM-DD 00:00:00` is the same as `YYYY-MM-DDT00:00:00`. 
> 
> ```python
> In [1]: import numpy as np
> In [2]: date_1 = np.datetime64('2025-05-20 00:00:00')
> In [3]: date_2 = np.datetime64('2025-05-20T00:00:00')
> In [4]: date_1 == date_2
> Out[4]: np.True
> ```
> So I will modify the code to use the `YYYY-MM-DD 00:00:00` format explicitly, but the remainder of the code does not need to change.
>  

 _Originally posted by @aysunrhn in [#164](https://github.com/sedtrails/sedtrails/issues/164#issuecomment-2897519373)_

manuGil:
> See my comment in issue #164, and confirm that this addresses the issue.
> 
> > In numpy, `YYYY-MM-DD 00:00:00` is the same as `YYYY-MM-DDT00:00:00`.
> > ```python
> > In [1]: import numpy as np
> > In [2]: date_1 = np.datetime64('2025-05-20 00:00:00')
> > In [3]: date_2 = np.datetime64('2025-05-20T00:00:00')
> > In [4]: date_1 == date_2
> > Out[4]: np.True
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So I will modify the code to use the `YYYY-MM-DD 00:00:00` format explicitly, but the remainder of the code does not need to change.
> 
> _Originally posted by @aysunrhn in [#164](https://github.com/sedtrails/sedtrails/issues/164#issuecomment-2897519373)_

It is OK if numpy treat accept both types of strings. What we have to care about it to enforce that in SedTRAILS we only accept one type. 
In practical terms, we should raise a custom exception when a different string is passed to the class.

src/sedtrails/particle_tracer/time.py
```diff
@@ -36,7 +36,7 @@ class Time:
 
     def _convert_to_datetime64(self, date_str: str) -> np.datetime64:
         """Convert reference_date in str format to numpy.datetime64"""
-        return np.datetime64(f"{date_str}T00:00:00", 's')
+        return np.datetime64(f"{date_str} 00:00:00", 's')
```
manuGil: Add an exception when the string does not match our CF format of choice. 
I recommend to create a custom exception for this. Shall I create that?

src/sedtrails/particle_tracer/time.py
```diff
@@ -47,23 +47,30 @@ def __post_init__(self):
...
-    def get_seconds_since_reference(self) -> int:
+    def get_seconds_since_reference(self) -> str:
         """
-        Returns the number of seconds since the reference date/time.
+        Returns the number of seconds since the reference date/time in a 
+        human readable format
         """
-        # Will complete once I can clarify the different between reference_date and simulation start
-        pass   
+        return str(self.get_current_time())
```
manuGil: I think I was misunderstood. I did not mean to say to modify this method. Rather, we need to add a new method for the human readable conversion, for example: `get_datetime(self)`, this should return the date and time (Gregorian time) in the same format as in our CF convention of choice. We should leave the get_seconds_since_reference() as it was, and return the number of seconds.

manuGil (changes requested):
Please, make the changes I mentioned.

src/sedtrails/particle_tracer/time.py
aysunrhn: OK. Then the `get_seconds_since_reference` should retrieve the simulation time in seconds (int): simply summing up the start_time and delta_seconds. So it should be just like `get_current_time` but without taking the `reference_date` into account

manuGil: I will have read the full code on the file, so I need a few minutes. But, with the changes we discussed today, we aren't redefining the methods (what they should do), only making changing the implementation. I will get back with concrete examples.

src/sedtrails/particle_tracer/time.py
aysunrhn: I created a custom exception inheriting from ValueError. Should we inherit from the `SedtrailsException` class instead? Which one is the better practice?

Also, I used regex to check for the reference_date format.

PR #173, opened 2025-05-21
Single particle simulation and Numba implementation

bartvanwesten:
# Single particle simulation and Numba implementation

## Overview
This PR introduces a first implementation of a Numba-accelerated particle tracing system along with example scripts demonstrating both the original and Numba pipeline for a single particle movement. The main goal is to establish the initial pipeline for particle simulation in SedTRAILS and demonstrate the performance benefits of using Numba.

### New Files:
- **position_calculator_numba.py**: Numba-optimized version of the particle position calculator
  - Implements both sequential and parallel modes (https://numba.pydata.org/numba-doc/dev/user/parallel.html)

### Example Scripts:
- **particle_simulation_example.py**: Basic single-particle simulation workflow
  - Demonstrates full flow field loading and particle tracking
  - Configured with specific parameters (X=40000, Y=17000, 6.333 hours duration, based on Matlab reference)
  - Includes visualization and performance metrics

- **particle_simulation_numba_example.py**: Performance comparison between implementations
  - Side-by-side comparison of original vs. Numba-optimized calculators
  - Tracks speed differences and calculation accuracy
  - Visualizes both trajectories for result verification

- **visualization_utils.py**: Utility functions for data visualization
  - Flow field plotting with vector visualization
  - Particle trajectory visualization functions

## Performance Benefits
Initial tests show significant performance improvements with the Numba implementation (up to 300x) while maintaining calculation accuracy between implementations (although some small differences).

## Implementation Status
This is a draft implementation that establishes the core functionality but requires further refinement. The Numba-specific code in particular is in a preliminary state and happens to work, but has not been fully investigated nor tested. The example scripts need to be better integrated with existing SedTRAILS data structures and functionalities, like receiving input from the YAML, using existing plotting functionality instead of temporary utils, properly seed particles using a seeder and store the trajectories in the correct data structure. @jreyns @manuGil  Could you pick this up?


manuGil:
Thank you Bart. I will pick this up.

manuGil (approved):
Thanks Bart. Given that these are examples that we will later use to connect the components in SedTRAILS. We can merge them without extensive review. Could you do so?

manuGil:
I am merging this pull request.

PR #176, opened 2025-05-26
Update Time class and the corresponding tests

aysunrhn:
Updated the Time class to follow the examples @bartvanwesten shared in [`examples/particle_simulation_example.py`](https://github.com/sedtrails/sedtrails/blob/dev/examples/particle_simulation_example.py)

Relevant files:
- `src/sedtrails/particle_tracer/time.py`
  - Added `duration` and `time_step` attributes to the Time class.
  - The new `end_time` property is calculated based on the `start_time` and the `duration` attributes in seconds
  - The `get_current_time` method now calculates the current time based on `start_time`, `step` and `time_step` in simulation
- `tests/particle_tracer/test_time.py`
  - Added new test functions to test `get_current_time` using simulation steps

### DIFF
--- a/src/sedtrails/particle_tracer/time.py
+++ b/src/sedtrails/particle_tracer/time.py
@@ -15,11 +15,11 @@ class Time:
     Attributes
     ----------
     reference_date : str
-        The reference date as string in format 'YYYY-MM-DD'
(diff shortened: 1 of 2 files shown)


manuGil (approved):
Thank you for these charges. They look on the right direction.  We might need to make small changes when  after coupling other components.

PR #181, opened 2025-05-28
Update Time class 

### LINKED ISSUE #180: Update Time class following changes to the simulation config parameters
Time class needs to be updated following the changes in the simulation config parameters as we discussed

aysunrhn:
Following the changes in the simulation configuration parameters, I updated the Time class in `src/sedtrails/particle_tracer/time.py`.

- Relevant files:
    - `src/sedtrails/particle_tracer/time.py`
        - `reference_date` and `start_time` are numpy.datetime64 objects. Input as str in 'YYYY-MM-DD hh:mm:ss' format
        - `duration` and `time_step` are numpy.timedelta64 objects. Input as str in human friendly ISO format. e.g. '3D 2H1M3S'
        - Added a static method to parse duration in str to seconds
        - Removed methods that are not in use at the moment
   - `tests/particle_tracer/test_time.py`
        - Updated tests to comply with changes in the Time class
        - Removed unnecessary tests
    - `src/sedtrails/exceptions/exceptions.py`
        - Defined a new custom exception `DurationFormatError` to catch incorrect duration input formats

Closes #180

### DIFF
--- a/src/sedtrails/particle_tracer/time.py
+++ b/src/sedtrails/particle_tracer/time.py
@@ -5,7 +5,7 @@ Time class used internally to represent simulation time.
 from dataclasses import dataclass, field
 import numpy as np
 import re
-from sedtrails.exceptions.exceptions import DateFormatError
+from sedtrails.exceptions.exceptions import DateFormatError, DurationFormatError
 
 @dataclass
 class Time:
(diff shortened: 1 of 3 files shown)


src/sedtrails/particle_tracer/time.py
```diff
@@ -31,99 +31,96 @@ class Time:
...
 
-    # reference_date should be a str, the class will do the transformation into a numpy.datetime64 object
+    # reference_date and start_time should be a str, 
+    # the class will do the transformation into a numpy.datetime64 object
     reference_date: str = field(default='1970-01-01 00:00:00')
-    start_time: int = 0
-    duration: int = 0
-    time_step: float = 1.0
+    start_time: str = field(default='1970-01-01 00:00:00')
+    duration: str = field(default='3D 2H1M3S')
```
manuGil, line 38: I think these defaults are misleading. I used those as an example, but I wasn't expecting they will make it to the code.

manuGil (approved):
This can be merged

PR #185, opened 2025-06-04
Modify Sphinx documentation action

### LINKED ISSUE #184: Update documentation rendering action
The action that runs documentation rendering should be updated as follows:

1. It should be triggered only by pull requests on the `main` branch
2. If possible, it should run only on files in the `docs` folder

aysunrhn:
Modifed the Sphinx documentation action so that the workflow runs only on the pull requests on the `main` branch when there has been changes made in contents of the `docs` folder

Closes #184

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,6 +4,8 @@ on:
     pull_request:
       branches:
         - main
+      paths:
+        - 'docs/**'
 
 permissions:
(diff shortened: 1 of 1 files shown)


manuGil (approved):
Thanks. It was straight forward, I see

manuGil:
After merging, could you test the workflow works as expected?

PR #187, opened 2025-06-11
Particle fluid data

aysunrhn:
I am trying to complete the particle and fluid data structures with all the necessary attributes and method. I suggest we have a PhysicalProperties class we can use in the particle classes. Also added tests for the new class.

Relevant files:
- `src/sedtrails/particle_tracer/particle.py`
- `tests/particle_tracer/test_particle.py`

### DIFF
--- a/src/sedtrails/particle_tracer/particle.py
+++ b/src/sedtrails/particle_tracer/particle.py
@@ -104,6 +104,29 @@ class Particle(ABC):
         """
         pass
 
+@dataclass
+class PhysicalProperties:
+    """
+    Base class for particle physical properties.
+    
+    Attributes
+    ----------
+    density : float
+        Material density in kg/m³
+    diameter : float
+        Particle diameter in meters
+    """
+    density: float
+    diameter: float
+
+
+    def __post_init__(self):
+        """Validate physical property values."""
+        if not isinstance(self.density, (int, float)) or self.density <= 0:
+            raise ValueError(f"Density must be positive, got {self.density}")
+        if not isinstance(self.diameter, (int, float)) or self.diameter <= 0:
+            raise ValueError(f"Diameter must be positive, got {self.diameter}")
+
 
 @dataclass
 class Mud(Particle):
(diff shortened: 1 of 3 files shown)


sgpearson17 (changes requested):
Thanks Aysun! This looks ok, but a few minor clarifications:
1. I think ``class Sand(Particle)`` should move up above ``class Mud(Particle)`` since we do not currently have other mud stuff implemented, and the main focus is currently on sand
2. there is some inconsistency in the terms used within ``class InterpolatedValue`` in the docstring vs where the floats are defined below. I know that's not the focus of this pull request but maybe it's good to change in this step. Not sure if we need to modify things elsewhere e.g. in the internal file structure too then?

- I specifically propose removing ``no_fractions``, as even if we have multiple "sediment fractions" (i.e., grain sizes) in a single simulation, each particle will only have one fraction. 
- I also propose modifying ``averaged_velocity``/``averaged_flow_velocity`` to ``depavg_flow_velocity``, as the averaging over depth (as opposed to time or another spatial dimension) is an important distinction. 
- Lastly, I think we should add ``_transport`` to ``bed_load`` and change ``suspended_sed_transport`` to ``suspended_load_transport`` to make them more consistent with each other

I have indicated these things below with an arrow -->
```
	bed_level : float
        The bed level of the particle in meters.
    --> depavg_flow_velocity : float
    -->      The depth-averaged flow velocity of the particle in m/s.
    ~~no_fractions : int~~
        ~~The number of fractions of the particle.~~
     --> bed_load_transport : ndarray
        The bed load sediment transport of the particle in kg/m/s.
     --> suspended_load_transport : float
     -->     The suspended loadsediment transport of the particle in kg/m/s.
    depth : float
        The water depth of the particle in meters.
    mean_shear_stress : float
        The mean bed shear stress of the particle in Pa.
    max_shear_stress : float
        The maximum bed shear stress of the particle in Pa.
    sed_concentration : float
        The suspended sediment concentration of the particle in kg/m^3.
    wave_velocity : ndarray
        The non-linear wave velocity of the particle in m/s.

```

Hope this is clear! happy to discuss!

aysunrhn:
Updates on this PR that we should discuss as a group:
- I updated the attributes in InterpolatedValue to be consistent with the SedtrailsData class in `src/sedtrails/transport_converter/format_converter.py`. I think we should establish a common naming convention for the variables throughout the repository.
- We should link the Fluid class with the Particle class through the InterpolatedValue.

I have opened issue #194 for the first topic.

sgpearson17 (approved):
approved as discussed

PR #197, opened 2025-06-12
Add physics converter for sediment transport calculations

bartvanwesten:
Implements a physics conversion toolbox that extents flow field data with sediment transport velocities and physical parameters. The system converts hydrodynamic and transport data into sediment transport velocities and equivalent transport thicknesses.

### 1. Physics Library (`physics_lib.py`)
- **Core physics calculations**, mainly following van Westen et al. (2025)
- **Modular functions** for:
  - Shear velocity computation from bed shear stress
  - Shields parameter (dimensionless bed shear stress) calculation
  - Bed load velocity using Soulsby equation with critical threshold detection
  - Suspended sediment velocity using complex ratio formulation with Rouse parameter
  - Transport layer thickness calculations for both bed load and suspended layers
  - Mixing layer thickness using Bertin (2008) method
  - Direction component derivation from transport magnitude and direction
  - Complete grain property calculations (dimensionless grain size, critical Shields, settling velocity)

### 2. Physics Converter (`physics_converter.py`)
- **PhysicsConverter class** that dynamically adds physics fields to existing SedtrailsData objects
- **PhysicsConfig dataclass** with default parameters:
  - Physical constants (gravity, von Kármán constant, viscosity)
  - Water/sediment properties (densities, porosity)
  - Grain characteristics (250 μm diameter for sand)
  - Method selection for different calculation approaches
- **Method-specific implementations**:
  - `van_westen_2025`: Complete workflow from van Westen et al. (2025)
  - `soulsby_2011`: Placeholder for alternative Soulsby approach

### 3. Enhanced SedtrailsData Structure
- **Dynamic physics field system** allowing runtime addition of computed parameters
- **New methods**: `add_physics_field()`, `has_physics_field()`, `get_physics_fields()`
- **Backward compatibility** with existing data access patterns
- **Automatic inclusion** of physics fields in time-indexed data retrieval

## Physics Fields Added
When physics conversion is applied, the following fields become available:

### Scalar Fields (shape: [time, fractions, spatial])
- `shields_number`: Dimensionless bed shear stress parameter
- `bed_load_layer_thickness`: Representative thickness of bed load transport layer [m]
- `suspended_layer_thickness`: Representative thickness of suspended transport layer [m]
- `mixing_layer_thickness`: Active mixing layer thickness following Bertin (2008) [m]

### Vector Fields (shape: [time, fractions, spatial] with 'x', 'y', 'magnitude' components)
- `bed_load_velocity`: Bed load sediment velocity [m/s]
- `suspended_velocity`: Suspended sediment velocity [m/s]

## Files Modified/Added
- `sedtrails/transport_converter/physics_lib.py` - Core physics calculation library
- `sedtrails/transport_converter/physics_converter.py` - Main converter class and configuration
- `sedtrails/transport_converter/format_converter.py` - Enhanced SedtrailsData with dynamic fields
- `examples/particle_simulation_numba_example.py` - Demonstration of physics-enhanced particle tracking

### DIFF
--- a/examples/particle_simulation_numba_example.py
+++ b/examples/particle_simulation_numba_example.py
@@ -16,10 +16,9 @@ import matplotlib.pyplot as plt
 
 # Import SedTRAILS modules
 from sedtrails.transport_converter.format_converter import FormatConverter, InputType
+from sedtrails.transport_converter.physics_converter import PhysicsConverter, PhysicsMethod
 from sedtrails.particle_tracer.data_retriever import FlowFieldDataRetriever
 from sedtrails.particle_tracer.particle import Sand
-
-# Import both calculator implementations
 from sedtrails.particle_tracer.position_calculator import ParticlePositionCalculator
 from sedtrails.particle_tracer.position_calculator_numba import create_numba_particle_calculator
 
@@ -61,11 +60,16 @@ print(f'Total timestamps: {time_info["num_times"]}')
 sedtrails_data = converter.convert_to_sedtrails_data()
 print(f'Data conversion completed in {time.time() - start_time:.2f} seconds')
 
+# Add physics calculations to the SedtrailsData
+physics_converter = PhysicsConverter()
+physics_converter.add_physics_to_sedtrails_data(sedtrails_data, method=PhysicsMethod.VAN_WESTEN_2025)
+
 # ===== STEP 2: Initialize Particle and Position Calculator =====
 print('\n=== STEP 2: Initializing Particle and Flow Field ===')
 
 # Create flow field data retriever
 retriever = FlowFieldDataRetriever(sedtrails_data)
+retriever.flow_field_name = "suspended_velocity" # Options: "depth_avg_flow_velocity", "bed_load_velocity", "suspended_velocity"
 
 # Get the initial flow field at specified timestep
 initial_time = sedtrails_data.times[TIMESTEP_INDEX]
@@ -91,6 +95,7 @@ simulation_start = time.time()
 current_time = initial_time
 
 for step in range(1, NUM_STEPS + 1):
+
     # Update current time
     current_time = initial_time + step * TIMESTEP
 
--- a/src/sedtrails/particle_tracer/data_retriever.py
+++ b/src/sedtrails/particle_tracer/data_retriever.py
@@ -122,10 +122,33 @@ class FlowFieldDataRetriever:
         # Get indices for interpolation
         lower_index, upper_index, weight = self.get_interpolation_indices(time)
         
+        # Helper function to handle fraction dimension
+        def extract_fraction(flow_data):
+            # Check the shape of the x component to determine fraction handling
+            x_data = flow_data['x']
+            if len(x_data.shape) == 1:
+                # Shape: (N) - no fraction dimension
+                return flow_data
+            elif x_data.shape[0] == 1:
+                # Shape: (1, N) - single fraction, squeeze out the fraction dimension
+                return {
+                    'x': flow_data['x'][0],
+                    'y': flow_data['y'][0], 
+                    'magnitude': flow_data['magnitude'][0]
+                }
+            else:
+                # Shape: (>1, N) - multiple fractions, select the specified fraction
+                return {
+                    'x': flow_data['x'][self.fraction_index],
+                    'y': flow_data['y'][self.fraction_index],
+                    'magnitude': flow_data['magnitude'][self.fraction_index]
+                }
+            
         # If time is exactly at a time step or outside the range, no interpolation needed
         if lower_index == upper_index:
             time_slice = self.sedtrails_data[lower_index]
             flow_field = time_slice[self.flow_field_name]
+            flow_field = extract_fraction(flow_field)
             
             return {
                 'x': self.sedtrails_data.x,
@@ -141,6 +164,8 @@ class FlowFieldDataRetriever:
             
             lower_flow = lower_slice[self.flow_field_name]
             upper_flow = upper_slice[self.flow_field_name]
+            lower_flow = extract_fraction(lower_flow)
+            upper_flow = extract_fraction(upper_flow)
             
             # Use the interpolation function for velocity components
             flow_x = self._interpolate_linearly(lower_flow['x'], upper_flow['x'], weight)
--- a/src/sedtrails/transport_converter/format_converter.py
+++ b/src/sedtrails/transport_converter/format_converter.py
@@ -29,6 +29,85 @@ class SedtrailsData:
     A data class for internally structuring SedTrails data.
     
     This class holds data for multiple time steps with time as the first dimension
+    for time-dependent variables. Physics fields can be added dynamically.
+    """
+    
+    times: np.ndarray
+    reference_date: np.datetime64
+    x: np.ndarray
+    y: np.ndarray
+    bed_level: np.ndarray
+    depth_avg_flow_velocity: Dict[str, np.ndarray]
+    fractions: int
+    bed_load_transport: Dict[str, np.ndarray]
+    suspended_transport: Dict[str, np.ndarray]
+    water_depth: np.ndarray
+    mean_bed_shear_stress: np.ndarray
+    max_bed_shear_stress: np.ndarray
+    sediment_concentration: np.ndarray
+    nonlinear_wave_velocity: Dict[str, np.ndarray]
+    
+    def __post_init__(self):
+        """Initialize container for dynamic physics fields."""
+        # Create a container for physics data that can be added later
+        self._physics_fields = {}
+    
+    def add_physics_field(self, name: str, data):
+        """
+        Add a physics field to the data structure.
+        
+        Parameters:
+        -----------
+        name : str
+            Name of the physics field
+        data : np.ndarray or dict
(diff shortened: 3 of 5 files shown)


src/sedtrails/particle_tracer/data_retriever.py
```diff
@@ -122,10 +122,33 @@ def get_flow_field(self, time: float) -> Dict[str, np.ndarray]:
...
+                    'magnitude': flow_data['magnitude'][0]
+                }
+            else:
+                # Shape: (>1, N) - multiple fractions, select the specified fraction
+                return {
+                    'x': flow_data['x'][self.fraction_index],
+                    'y': flow_data['y'][self.fraction_index],
+                    'magnitude': flow_data['magnitude'][self.fraction_index]
+                }
+
```
manuGil, line 146: we should discuss this with the team because @sgpearson17 mentioned that representing 'fractions' wasn't necessary a while ago. But, they are necessary here.

examples/particle_simulation_numba_example.py
```diff
@@ -61,11 +60,16 @@
 sedtrails_data = converter.convert_to_sedtrails_data()
 print(f'Data conversion completed in {time.time() - start_time:.2f} seconds')
 
+# Add physics calculations to the SedtrailsData
+physics_converter = PhysicsConverter()
+physics_converter.add_physics_to_sedtrails_data(sedtrails_data, method=PhysicsMethod.VAN_WESTEN_2025)
```
manuGil, line 65: Let's try to keep the names of the functions as short as possible. A way to do it is to avoid mentioning the context on the name unless it is necessary because clarity must be given. For example, if the name `add_physics_to_sedtrails_data` would be only necessary if there is also a function called `add_physics_to_another_data` or the like; otherwise, something like `add_physics` would be enough. The context and details should go in the docstrings.

src/sedtrails/transport_converter/physics_converter.py
```diff
@@ -1,27 +1,271 @@
...
+        if chosen_method == PhysicsMethod.VAN_WESTEN_2025:
+            self._add_physics_van_westen(sedtrails_data)
 
+        elif chosen_method == PhysicsMethod.SOULSBY_2011:
+            self._add_physics_soulsby(sedtrails_data)
 
-PHYSICS_PLUGINS = "plugins.physics.sand"
+        else:
+            raise ValueError(f"Unknown physics method: {chosen_method}. "
+                             f"Available methods: '{PhysicsMethod.VAN_WESTEN_2025}', '{PhysicsMethod.SOULSBY_2011}'")
```
manuGil, line 111: Early on the development, we decided to implement the choices for methods using the Python plugin approach. So that, new methods could be easily added laer on. There is an example of how to implement plugins in https://github.com/sedtrails/sedtrails/tree/dev/src/sedtrails/transport_converter/plugins Does your implementation also provide that flexibility?

src/sedtrails/transport_converter/physics_converter.py
```diff
@@ -1,27 +1,271 @@
...
+import numpy as np
+from typing import Optional
+from dataclasses import dataclass
+
+# Import physics library
+from sedtrails.transport_converter import physics_lib
+
+class PhysicsMethod:
+    VAN_WESTEN_2025 = "van_westen_2025"
+    SOULSBY_2011 = "soulsby_2011"
```
manuGil, line 23: Please check the idea of plugins as we thought to apply it in https://github.com/sedtrails/sedtrails/tree/dev/src/sedtrails/transport_converter/plugins

src/sedtrails/transport_converter/physics_converter.py
```diff
@@ -1,27 +1,271 @@
...
+                             f"Available methods: '{PhysicsMethod.VAN_WESTEN_2025}', '{PhysicsMethod.SOULSBY_2011}'")
 
-# import converter dynamically
-physics_converter = importlib.import_module(PHYSICS_PLUGINS, ".")
+        print(f"Physics calculations complete using {chosen_method} method!")
 
-# this guarantees that we can invoke the convertr
-# independent of its name
-converter = physics_converter.PhysicsPlugin()
+    def _add_physics_van_westen(self, sedtrails_data):
```
manuGil, line 115: Related to the plugin idea. If we implement methods as part of class, like in this case, then we will have to implement a new function for every new method we want, and change the code in several places, eg. the `PhysicsMethod`, the if-statements  to check which method has been chosen, etc. As a consequence it is more cumbersome to extend the library of methods that can be apply. Although, if there is a good reason to implement things this way, let's discuss it tomorrow.

src/sedtrails/transport_converter/physics_lib.py
```diff
@@ -1,3 +1,428 @@
 """
-Methods to convert between different units of physics parameters. 
+Physics Library for Sediment Transport Calculations
```
manuGil, line 2: In the physics library, I imagined  a list of independent functions which take the data we want to transform and returns the converted values. I will explain the idea on Wednesday.

src/sedtrails/transport_converter/format_converter.py
```diff
@@ -28,6 +28,85 @@ class SedtrailsData:
...
+        Returns:
+        --------
+        bool
+            True if any physics fields exist
+        """
+        return len(self._physics_fields) > 0
+    """
+    A data class for internally structuring SedTrails data.
+    
     This class holds data for multiple time steps with time as the first dimension
```
manuGil: Shouldn't all docstrings go on the top of the class? These look out of place. Can you check there are correctly placed?

src/sedtrails/transport_converter/physics_converter.py
```diff
@@ -1,27 +1,271 @@
...
+    kinematic_viscosity: float = 1.36e-6  # m^2/s (10°C, 35 ppt)
+    water_density: float = 1027.0  # kg/m^3 (10°C, 35 ppt)
+    particle_density: float = 2650.0  # kg/m^3 (quartz)
+    porosity: float = 0.4  # [-]
+    
+    # Sediment properties
+    grain_diameter: float = 2.5e-4  # m (250 μm)
+    
+    # Morphological acceleration factor
+    morfac: float = 1.0
```
manuGil, line 42: If this are constants across simulations, let's write them at global variables in the file, and in the same fashion as MAGIC_NUMBERS. If these are values that could be overwritten based on the configuration file, then, this code should treat them as local variables, and we read them thru the `configuration interface`. Let's discuss this with the goal of deciding on that.

src/sedtrails/transport_converter/physics_converter.py
```diff
@@ -1,27 +1,271 @@
...
+
 
-# type of physics plugin to use.
-## the value  of PHYSICS_PLUGINGS can be set based on the
-## modeler's choice in the configuration file.
+class PhysicsConverter:
+    """
+    Add physics calculations to existing SedtrailsData using different methods.
+    
+    Each method (van Westen, Soulsby, etc.) has its own complete workflow.
```
manuGil, line 56: We should consider if this coupling between method and workflow can be separated via the plugin idea.

manuGil (commented):
Let's discuss in today's meeting

src/sedtrails/particle_tracer/data_retriever.py
bartvanwesten: The fractions-dimension is included in the Delft3D-data that is loaded, so that's why I included it here. If "regular" simulation data is used (with 1 fraction), the users doesn't have to notice anything. But this implementaiton allows the user to select for which Delft3D fraction the sediment transport data is used. It does not directly relate to the Sedtrails particles themselves. @sgpearson17 Or do you have another prefered way of implementation?

examples/particle_simulation_numba_example.py
bartvanwesten: I agree. I was struggeling with the correct naming conventions. @manuGil Could you maybe do a suggestion for the names of all these functions in your "larger" overhaul of the code?

src/sedtrails/transport_converter/physics_converter.py
bartvanwesten: As discussed in the meeting, no it currently does not, but it should ;)

src/sedtrails/transport_converter/format_converter.py
bartvanwesten: Yes, changed it

src/sedtrails/transport_converter/physics_converter.py
bartvanwesten: Yes, these should all come from the configuration interface, but this was just coded in a "pragmatic" or lazy manner to get a working version.

src/sedtrails/particle_tracer/data_retriever.py
manuGil: I see. then, is this something we have to include in the configuration file? I remove it from there after discussing it with @sgpearson17 We can put it back anytime.
Let's continue the discussion in this issue #200

examples/particle_simulation_numba_example.py
manuGil: Sure I will to so.

src/sedtrails/transport_converter/physics_converter.py
manuGil: that's ok, I will keep it in mind when working in the general workflow. We should pass those variable via the configuration interface.

PR #205, opened 2025-06-18
Readme

Hassan-Shafiei:
Updates the installations instructions in the readme file.


README.md
```diff
@@ -12,7 +12,7 @@ SedTRAILS requires Python 3.12 and above, but it is likely to work with Python 3
 
 1. Clone the repository:
 ```bash
-git clone 
+git clone <your repository SSH key>
```
manuGil: Thanks for this Hassan. What goes in `<your repository SSH key>` it is not an SSH key. It is the path to the remote repository which can use SSH or HTTPS. Can you please substitute the that placeholder for  `https://github.com/sedtrails/sedtrails.git`? I think in this case providing a command that actually works would be best for the users. And the HTTPS path is best because it doesn't require setting an SSH connection.

manuGil (changes requested):
Please make the changes I suggest. Thank you for your contribution.

README.md
Hassan-Shafiei: I resolved the comment

manuGil (approved):
@Hassan-Shafiei Thanks

PR #204, opened 2025-06-18
Fix configuration schema files

aysunrhn:
This PR resolves the redundancies and in configuration schemas, and updates the `main.schema.json` file to be consistent.  

- Deleted [src/sedtrails/configuration_interface/config_schema.json](https://github.com/sedtrails/sedtrails/blob/dev/src/sedtrails/configuration_interface/config_schema.json)
- Updated [src/sedtrails/config/main.schema.json](https://github.com/sedtrails/sedtrails/blob/dev/src/sedtrails/configuration_interface/config_schema.json) to bec consistent with the current implementation

Closes #202

PR #208, opened 2025-06-21
Netcdf writer implementation

### LINKED ISSUE #203: Implement NetCDF writer in Data Manager module
We need to implement NetCDF to write simulation outputs.

### Relevant files
- [`src/sedtrails/data_manager/netcdf_writer.py`](https://github.com/sedtrails/sedtrails/blob/dev/src/sedtrails/data_manager/netcdf_writer.py)

aysunrhn:
### Changes

- Added the NetCDFWriter class to the `netcdf_writer.py` file to write SedTRAILS simulation output to a given path in NetCDF format.
  - NetCDFWriter is similar to the NetCDFReader class in `netcdf_reader.py`. We work with the `xugrid` package now, not `xarray`.
  - NetCDFWriter expects a xu.UgridDataset, filename and an output file path
  - We check to make sure the filename ends with the correct file extension
  - If the output path already exists, we create a new folder appending the timestamp to avoid overriding
  - The output file path will also store the image outputs from the visualization analysis in the future
- Added a test for the NetCDFWriter class in `tests/data_manager/test_netcdf_writer.py` 
  - We can add a test for the NetCDFReader class in the same directory in the future

Closes #203. 
Note: This PR should be merged after #204


tests/data_manager/test_netcdf_writer.py
```diff
@@ -0,0 +1,44 @@
+import pytest
+import xugrid as xu
+import numpy as np
+from pathlib import Path
```
manuGil, line 4: Ruff flagged this import as 'unused'. Could you remove it, please? After that you can merged the code to the dev branch

manuGil (approved):
Please, merge after addressing the comment. Thanks

tests/data_manager/test_netcdf_writer.py
aysunrhn: Interesting, Path is used in line 22, though. Ruff on my vscode does not flag it either for that reason.

manuGil: ok, then we can merge it

PR #211, opened 2025-06-25
Add xugrid to dependencies

### LINKED ISSUE #207: Add xugrid to the dependencies list
We decided to use `xugrid` instead of `xarray` in the project. In both the `netcdf_reader.py` and the `netcdf_writer.py` (new class I am working on right now) we use `xugrid` currently. We need to update the dependencies in the `environment.yml` and `pyproject.toml` files and add `xugrid` package.

### Relevant files
- [`environment.yml`](https://github.com/sedtrails/sedtrails/blob/dev/environment.yml)
- [`pyproject.toml`](https://github.com/aysunrhn/sedtrails/blob/main/pyproject.toml)

aysunrhn:
Add `xugrid` package to the list of dependencies in `environment.yml` and `pyproject.toml` files.

### Relevant files
- [`environment.yml`](https://github.com/sedtrails/sedtrails/blob/dev/environment.yml)
- [`pyproject.toml`](https://github.com/sedtrails/sedtrails/blob/dev/pyproject.toml)

Closes #207

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -17,4 +17,5 @@ dependencies:
   - sphinx-rtd-theme
   - pyyaml
   - jsonschema
+  - xugrid
   
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,6 +13,7 @@ dependencies = [
   "matplotlib",
   "xarray",
   "typer",
+  "xugrid"
 ]
 requires-python = ">=3.12"
 authors = [


aysunrhn:
Do we still need `xarray` in the dependencies? @manuGil

manuGil:
No, if uxgrid already has it as dependency.  Can you check if that's the case?

aysunrhn:
> No, if uxgrid already has it as dependency. Can you check if that's the case?

Yes. `xugrid` package has `xarray` as a dependency. I suggest we remove `xarray` from our dependencies then to keep it simple.

manuGil (approved):

PR #213, opened 2025-06-25
Update simulation buffer to connect NetCDF writer to the simulation loop

### LINKED ISSUE #212: Implement simulation buffer to store data
Implement simulation buffer in `simulation_buffer.py` which we can use in simulation loops, to pass simulation output to `netcdf_writer.py`

aysunrhn:
Implemented simulation buffer in `simulation_buffer.py` which we can call inside the simulation loop. It is consistent with the `netcdf_writer.py` file. I also added unit tests for the `simulation_buffer.py`.

Important points to note:
- Stores all relevant simulation data in memory, at the moment this is position (x and y), particle ID and time
- Allows us to add new data per time step/particle
- Can clear/reset for new runs (wrote a `clear` method for this)
- Exports to a `xu.UgridDataset` which is compatible with the `NetCDFWriter` class in `netcdf_writer.py`

I followed Bart's examples to try to get a sense of what we should store in the buffer. With the current implementation, we can call it as follows:
```python
buffer = SimulationDataBuffer()
for step in range(NUM_STEPS):
    buffer.add(particle_id, current_time, pos_x, pos_y)
```

### Relevant files
- `simulation_buffer.py`
- `test_simulation_buffer.py`

Closes #212 

PS: I realized the ruff error @manuGil pointed about the Path variable was in the `tests/data_manager/test_netcdf_writer.py` file but I did not notice it at the time. So I also fixed that in this PR.

### DIFF
--- a/src/sedtrails/data_manager/simulation_buffer.py
+++ b/src/sedtrails/data_manager/simulation_buffer.py
@@ -1,5 +1,96 @@
-"""
-Simulation Data Buffer
-======================
-Temporarily stores chunks of simulation data in memory, while they wait to be written to a simulation file.
-"""
+import xugrid as xu
+import numpy as np
+
+class SimulationDataBuffer:
+    """
+    Temporarily stores chunks of simulation data in memory, while they wait to be written to a simulation file.
+
+    Attributes
+    ----------
+    buffer : dict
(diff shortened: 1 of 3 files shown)


manuGil (approved):
As far I can tell, the code looks good. Thank you

manuGil:
I am merging this before working further on the integration of components.

PR #219, opened 2025-07-02
Implement memory manager

### LINKED ISSUE #217: Implement memory manager
We need to implement memory manager in the data manager module to enforce a memory limit on the simulation buffer. The limit should be 512MB, when the SimulationDataBuffer object exceeds this size, we write the buffer into file(s).

aysunrhn:
- Implemented memory manager in `memory_manager.py` as the MemoryManager class. 
    - This class checks the size of the SimulationDataBuffer object, and if it exceeds 512MB it calls the NetCDFWriter to write out the simulation results in chunks
    - This way we can avoid losing any simulation data in case the simulation crashes or ends prematurely
    - In the future, we can implement the simulation recovery module to restart simulation from where it ended
- MemoryManager class also includes a method to merge all output files (if there are multiple files, that is) 
    - A single file is more convenient for downstream analysis and it is easier to store, transfer and keep track of simulations
    - We assume that each chunk written out to the filesystem covers a specific timespan within the simulation, so the merging is done based on the "time" coordinates of the xugrid dataset
    - I modified the SimulationDataBuffer object to have "time" as both a coordinate and dimension to allow us to merge based on the time coordinate
    - We assume that time coordinate is non-overlapping between different chunks of files

I also implemented tests for the MemoryManager class in `tests/data_manager/test_memory_manager.py`, while doing so I noticed a shortcoming in the NetCDFWriter class:
-  I modified the class to make sure we have an XUGRID compliant data structure as output
-  When testing the MemoryManager, the output_dir variable is extracted from the NetCDFWriter class, this decouples the two components and makes our code modular

### Relevant files:
- `src/sedtrails/data_manager/memory_manager.py`
- `src/sedtrails/data_manager/simulation_buffer.py`
- `src/sedtrails/data_manager/netcdf_writer.py`
- `tests/data_manager/test_memory_manager.py`

Closes #217

### DIFF
--- a/src/sedtrails/data_manager/memory_manager.py
+++ b/src/sedtrails/data_manager/memory_manager.py
@@ -3,3 +3,84 @@ Memory Manager
 ==============
 Manages memory allocation and deallocation for the simulation data buffer.
 """
+
+import sys
+from sedtrails.data_manager.netcdf_writer import NetCDFWriter
+
+class MemoryManager:
+    """
+    Manages memory usage for the simulation data buffer. If the buffer exceeds the memory limit,
+    writes its contents to a NetCDF file and clears the buffer.
+
+    The current memory limit is set to 512 MB
+
+    Attributes
+    ----------
+    max_bytes : int
+        Maximum allowed memory usage in bytes.
+    writer : NetCDFWriter
+        Writer instance to write buffer to NetCDF when limit is exceeded.
+    file_counter : int
+        Counter for naming output files uniquely.
+    """
+
+    def __init__(self, output_dir, max_bytes=512 * 1024 * 1024):
+        """
+        Initialize the memory manager.
+
+        Parameters
+        ----------
+        output_dir : str or Path
+            Directory to write NetCDF files.
+        max_bytes : int
+            Maximum allowed memory usage in bytes (default: 512 MB).
+        """
+        self.max_bytes = max_bytes
+        self.writer = NetCDFWriter(output_dir)
+        self.file_counter = 0
+
+    def buffer_size_bytes(self, buffer):
+        """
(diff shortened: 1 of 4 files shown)


src/sedtrails/data_manager/memory_manager.py
```diff
@@ -2,4 +2,106 @@
...
+            The simulation data buffer instance from simulation_buffer.py.
+        node_x, node_y, face_node_connectivity, fill_value : arrays/values
+            Mesh info for NetCDF writing.
+        """
+        if self.buffer_size_bytes(sim_buffer.buffer) > self.max_bytes:
+            filename = f"sim_buffer_{self.file_counter}.nc"
+            ugrid_ds = sim_buffer.to_ugrid_dataset(node_x, node_y, face_node_connectivity, fill_value)
+            self.writer.write(ugrid_ds, filename)
+            sim_buffer.clear()
+            self.file_counter += 1
```
manuGil, line 87: This a question to confirm all cases are covered, no specifically about these code lines. In the case, the buffer doesn't overflows, we still have to guarantee that data gets written to a file by the netcdf_writer. Is that cover in the current implementation?

manuGil (approved):
If the case I mentioned in another comment is covered here or in existing code. Then, this is good to merge

src/sedtrails/data_manager/memory_manager.py
aysunrhn: NetCDFWriter class has a write method which will write the dataset to a file. But it needs to be called explicitly within the simulation.

This is also got me thinking about having a check to make sure this method is called at least one throughout the simulation.

manuGil: I will merge this to work on the integration.

PR #252, opened 2025-07-25
Implement exception logging 

### LINKED ISSUE #251: Implement global logging to catch all exceptions

aysunrhn:
Implemented an exception logging system for SedTRAILS that captures and logs exceptions throughout the simulation run. I added an example log file in the `examples` folder.

Closes #251 

The logger module has the features:
- Global Exception Handler: Automatically captures and logs unhandled exceptions
- Logger Manager: Centralized logging configuration with directory management
- Exception Logging Functions: log_exception() and log_simulation_state() for structured error reporting

I also added unit tests for individual logging tasks such as:
- Test core logging functionality, creating log files successfully
- Test that global exception handler works as intended, and they keyboard interruption (Ctrl+C) is ignored
- Test that logger is integrated with the Simulation class

## Relevant files
- `src/sedtrails/simulation.py`
- `tests/logger/test_logger.py`
- `examples/log_example.txt`

### DIFF
--- a/src/sedtrails/logger/logger.py
+++ b/src/sedtrails/logger/logger.py
@@ -60,196 +60,198 @@ class LoggerManager:
 
         return self.logger
 
-def _format_seeding_strategy(strategy):
-    """Helper function to format seeding strategy for readable logging."""
-    if not strategy or strategy == 'unknown':
-        return 'none'
-    
-    if isinstance(strategy, str):
-        return strategy
-    
-    if isinstance(strategy, dict):
-        formatted_parts = []
+    def log_simulation_state(self, state: dict, level=logging.INFO) -> None:
+        """
+        Logs the current state of the simulation with human-readable sentences.
+        Long messages are split into multiple lines for readability.
+        """    
+        status = state.get('status', state.get('state', 'unknown'))
         
-        for strategy_type, config in strategy.items():
-            if strategy_type == 'point':
-                points = config.get('points', [])
-                if isinstance(points, list):
-                    point_count = len(points)
-                    if point_count == 1:
-                        formatted_parts.append(f"single point at {points[0]}")
+        # Create messages, split when necessary
+        if status == 'simulation_started':
+            command = state.get('command', 'unknown')
+            config = state.get('config_file', 'unknown')
+            python_ver = state.get('python_version', 'unknown')
+            working_dir = state.get('working_directory', 'unknown')
+            
+            # Split into multiple log lines
+            self.logger.log(level, "=== SIMULATION START ===")
+            self.logger.log(level, f"Command: {command}")
+            self.logger.log(level, f"Python: {python_ver}")
+            self.logger.log(level, f"Config: {config}")
+            self.logger.log(level, f"Working directory: {working_dir}")
+            
+        elif status == 'config_loading':
+            config_path = state.get('config_file_path', 'unknown')
+            self.logger.log(level, f"Loading configuration: {config_path}")
+            
+        elif status == 'simulation_parameters':
+            duration = state.get('duration', 'unknown')
+            timestep = state.get('timestep', 'unknown')
+            start_time = state.get('start_time', 'unknown')
+            strategy = state.get('seeding_strategy', 'unknown')
+            output_dir = state.get('output_dir', 'unknown')
+            
+            # Format seeding strategy for readability
+            formatted_strategy = self._format_seeding_strategy(strategy)
+            
+            # Split parameters into multiple lines
+            self.logger.log(level, "--- Simulation Parameters ---")
+            self.logger.log(level, f"Duration: {duration}")
+            self.logger.log(level, f"Time step: {timestep}")
+            self.logger.log(level, f"Start time: {start_time}")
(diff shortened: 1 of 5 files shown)


src/sedtrails/simulation.py
```diff
@@ -12,17 +12,31 @@
...
 
+def setup_global_exception_logging():
+    """Setup global exception logging for unhandled exceptions."""
+    original_excepthook = sys.excepthook
+    
+    def exception_handler(exc_type, exc_value, exc_traceback):
+        # Don't log KeyboardInterrupt (Ctrl+C)
+        if issubclass(exc_type, KeyboardInterrupt):
+            original_excepthook(exc_type, exc_value, exc_traceback)
+            return
```
manuGil, line 27: Do you think we need this in for our use case? I think we want to log the interruptions, but maybe you have a different idea.

src/sedtrails/simulation.py
```diff
@@ -12,17 +12,31 @@
 from sedtrails.data_manager import DataManager
 from sedtrails.particle_tracer.timer import Time, Duration, Timer
 from sedtrails.logger.logger import log_simulation_state, log_exception, _logger_manager
```
manuGil: my understanding is that `_variable_name` is a convention to indicate private methods/attributes in Python. But here, we are treading `_logger_manager` as public. Could you please double check the code in logger.py is being used as intended? 
Besides, I would expect that we only need to import a single object from logger, and use that thru the script.

src/sedtrails/simulation.py
```diff
@@ -179,23 +204,28 @@ def run(self):
 
         from tqdm import tqdm
 
+        if not self._config_is_read:  # assure config is read only once
+            self._controller.load_config(self._config_file)
+            self._config_is_read = True
+
+            # Ensure logger directory is set from config
+            output_dir = self._controller.get('folder_settings.output_dir', 'results')
+            _logger_manager.log_dir = output_dir
```
manuGil: This is also in line 70. Why do we need it here as well?

src/sedtrails/logger/logger.py
manuGil, line 1: let's have a discussion on your implementation of this. I do not understand some of the codding conventions you used.

tests/logger/test_logger.py
```diff
@@ -0,0 +1,523 @@
...
+import shutil
+import sys
+import logging
+
+from sedtrails.simulation import Simulation, setup_global_exception_logging
+from sedtrails.logger.logger import _logger_manager, log_exception, log_simulation_state
+from sedtrails.exceptions.exceptions import ConfigurationError
+
+# Test Configuration Data
+def get_valid_config(output_dir):
```
manuGil: Unit-test for the logger shouldn't be testing the validity of the configuration parameters. This has already been tested by the module dealing with the configuration. If you use a code assistant, please do not let it make all the work. The output should always be check for consistency and logic

manuGil (changes requested):
Hello Aysun, I can see that you have implemented a global way to log exceptions. However, I have many doubts regarding the way the logger is implemented. It looks like the code and tests aren't align with the development goals. 
Let's have a discussion about this before you make more changes.

aysunrhn:
UPDATE: following our last discussion about the logging module

- .Added `log_simulation_state()` and ` log_exception()` methods to the `LoggerManager` class in the `logger.py` file. Moved them from standalone functions into the LoggerManager class itself. This simplifies how we use the LoggerManager
```python
# Create manager and set up logger
logger_manager = LoggerManager(output_dir)
logger_manager.setup_logger()

# Use manager methods directly
logger_manager.log_simulation_state(state)
logger_manager.log_exception(e, "context")
```
- Updated `simulation.py` and `test_logger.py` to be consistent with the changes in the `logger.py`

src/sedtrails/simulation.py
manuGil: @aysunrhn was this possible? If not that's ok, we can discuss with the team if this is useful or not

tests/logger/test_logger.py
```diff
@@ -0,0 +1,222 @@
...
+        """Test that logger raises error if directory does not exist."""
+        non_existent_dir = os.path.join(self.temp_dir, 'does_not_exist')
+        # Ensure the directory does not exist
+        if os.path.exists(non_existent_dir):
+            shutil.rmtree(non_existent_dir)
+        logger_manager = LoggerManager(non_existent_dir)
+        with pytest.raises(FileNotFoundError):
+            logger_manager.setup_logger()
+
+if __name__ == '__main__':
```
manuGil: Why do we need to call the script itself from main. Pytest automatically discover test files, classes and functions. Unless, this is an special case. I suggest to remove this

manuGil (commented):
@aysunrhn  Could you please read my new comments and questions, after that you can merge this. In the upcoming tasks that may involve `simulation.py`, it would be best not to modify that file. I am still working on it and when we both make changes, we have to solve conflicts. I suggest, we test other components in separated files, and we leave the integration (use) to a bit later (once I have completed the work on `simulation.py`).

PR #262, opened 2025-08-06
stuart's experiments with adding documentation

sgpearson17:
attempted to add some introductory material and edit md/rst files. I also pasted in the command line stuff that we went through. I have been looking through this (https://coderefinery.github.io/documentation/sphinx/#exercise-sphinx-quickstart and https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#ensuring-the-code-can-be-imported) and would like to figure out how I can get the autodoc stuff working but I don't want to break anything


manuGil:
@sgpearson17 I see a few advantages of using `autodoc2`: we can directly include python packages and files as pages of the documentation. However, I also see some disadvantages, excluding part of the code that we do not want to include must require that we filter functions and class out, on the `__init__.py` file of each package. 
It is difficult to see which one will give us less work. But we can try `autodoc2` and see where we go from the. 

@aysunrhn, if Stuart want to try `autodoc2` Could you help Stuart to substitute the 'autodoc' extension in Sphinx for the `autodoc2` and complete the set up?

docs/sourcecode/particle-tracer.rst
```diff
@@ -0,0 +1,89 @@
...
+===================
+
+This is an overview of the SedTRAILS particle tracer, which is designed to simulate the transport and dispersion of sediment particles in aquatic environments. The particle tracer is implemented as a Python package, and the source code is available in the `sedtrails.particle_tracer` module. 
+
+Data Retriever
+--------------
+The data retriever module is responsible for fetching and processing data required for particle tracking simulations. It includes functions to retrieve hydrodynamic data, sediment properties, and other relevant information needed for accurate particle transport modeling.
+
+.. automodule:: sedtrails.particle_tracer.data_retriever
+                    :members:
```
manuGil, line 11: Did this build without errors? RST is very strict. Spaces and alignment matter. For example, in this line, the compilation might be failing because `:member:` must start exactly under the `a` in `automodule`.

docs/sourcecode/particle-tracer.rst
```diff
@@ -0,0 +1,89 @@
...
+
+
+Diffusion Library
+----------------
```
manuGil, line 15: Also for marking sections and subsections, underscoring must be at least as long as the text for the section. This is, underscoring can not undershoot, but it can overshoot. I this case, one `-` is missing 🫤

manuGil (commented):
Thanks for documenting the code. Please consider my comments regarding the use of RST.
If you want to try `autodoc2` let us know and we can help you to set it up. 

P.S. Do not be afraid to break things up, it ti comes to the worst, we have version control and always could roll (most of) things back 😆

sgpearson17:
Hi @manuGil and @aysunrhn , I think indeed implementing autodoc2 could be nice- if that is the case, should I stop trying to set up everything in \docs\sourcecode ?

manuGil:
> Hi @manuGil and @aysunrhn , I think indeed implementing autodoc2 could be nice- if that is the case, should I stop trying to set up everything in \docs\sourcecode ?

@aysunrhn Could you follow with Stuart to try that out? I will merge this PR so that we can compare the tools once autodoc2 is setup

manuGil (approved):

PR #264, opened 2025-08-06
Implement integrations test for physics plugins

### LINKED ISSUE #253: Create integration test for plugins
As discussed when looking at #246 , we need to create an integration test to check the validity of new plugins that we (or external users implement). This came up because we want to test Natascia's Soulsby implementation in the model.

aysunrhn:
Implemented integration tests for the physics plugin system in `src/sedtrails/transport_converter/plugins/physics/`  to ensure new plugins are compliant with the existing structure.

Closes #253 

### Relevant files:
- `tests/transport_converter/plugins/test_physics_plugin.py`:
    - Each plugin inherits from the BasePhysicsPlugin class.
    - Each plugin has an `add_physics` method.
    - Each plugin's `add_physics` method is callable.
    - Each plugin has `sedtrails_data` as parameter. 

- `pyproject.toml`
    - I added a custom marker (called "integration") to mark these tests, so that we can run them separately from the unit tests using `pytest -m integration`.

- `.github/workflows/pytest.yml`
   - Split the pytest action into 2 jobs: unit test and integration test
   - Integration job needs the unit job to run first

### DIFF
--- a/tests/transport_converter/plugins/test_physics_plugin.py
+++ b/tests/transport_converter/plugins/test_physics_plugin.py
@@ -60,4 +60,22 @@ def test_plugin_add_physics_callable(plugin_class):
     method = getattr(plugin_class, "add_physics", None)
     assert callable(method), (
         f"{plugin_class.__name__}.add_physics is not callable"
-    )    
+    )
+
+@pytest.mark.integration
+@pytest.mark.parametrize("plugin_class", get_plugin_classes())
+def test_plugin_add_physics_accepts_sedtrails_data(plugin_class):
+    """
+    Integration test to ensure each plugin's add_physics method accepts sedtrails_data as parameter.
+    """
+    method = getattr(plugin_class, "add_physics", None)
+    assert method is not None, f"{plugin_class.__name__} does not have an add_physics method"
+    
+    sig = inspect.signature(method)
+    param_names = list(sig.parameters.keys())
+    
+    # Should we check the position of the sedtrails_data as well?
+    assert "sedtrails_data" in param_names, (
+        f"{plugin_class.__name__}.add_physics should have 'sedtrails_data' parameter. "
+        f"Found parameters: {param_names}"
+    )
--- a/tests/transport_converter/plugins/test_physics_plugin.py
+++ b/tests/transport_converter/plugins/test_physics_plugin.py
@@ -72,10 +72,12 @@ def test_plugin_add_physics_accepts_sedtrails_data(plugin_class):
     assert method is not None, f"{plugin_class.__name__} does not have an add_physics method"
     
     sig = inspect.signature(method)
-    param_names = list(sig.parameters.keys())
+    params = list(sig.parameters.values())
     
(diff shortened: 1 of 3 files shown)


tests/transport_converter/plugins/test_physics_plugin.py
```diff
@@ -0,0 +1,81 @@
...
+    method = getattr(plugin_class, "add_physics", None)
+    assert method is not None, f"{plugin_class.__name__} does not have an add_physics method"
+    
+    sig = inspect.signature(method)
+    param_names = list(sig.parameters.keys())
+    
+    # Should we check the position of the sedtrails_data as well?
+    assert "sedtrails_data" in param_names, (
+        f"{plugin_class.__name__}.add_physics should have 'sedtrails_data' parameter. "
+        f"Found parameters: {param_names}"
```
manuGil: I think we should check for the order. In the base-plugin it must the the first parameter

```python
    def add_physics(self, sedtrails_data: SedtrailsData, *args, **kwargs):
 ```

File: tests/transport_converter/plugins/test_physics_plugin.py [hunk 2]
```diff
@@ -0,0 +1,81 @@
+import os
+import importlib.util
+import inspect
+import pytest
+from sedtrails.transport_converter.plugins.physics.plugin import BasePhysicsPlugin
+
+# TODO: Is there a better way to get the plugin directory?
```
manuGil, line 7: I think the path can be hard-coded because this will only run on the source code. Can you try if tests still work if we hard-code the path to the plugin directory, both when running locally and on GitHub?

manuGil (approved):
Thanks for this Aysun. please consider my comments before merging.

manuGil:
I just noticed that the integration tests failed. Can you fix those please? I also recommend to print the name of the plugin that fails as part of the assertion message. that way it will be easier to identify the 'offender' plugin.

aysunrhn:
> I just noticed that the integration tests failed. Can you fix those please? I also recommend to print the name of the plugin that fails as part of the assertion message. that way it will be easier to identify the 'offender' plugin.

Yes it fails because `soulsby.py` does not have the sedtrails_data as input. It is the latest test that I added for the plugins. Should I modify the `soulsby.py` script? It is a placeholder right now anyways.

manuGil:
> > I just noticed that the integration tests failed. Can you fix those please? I also recommend to print the name of the plugin that fails as part of the assertion message. that way it will be easier to identify the 'offender' plugin.
> 
> Yes it fails because `soulsby.py` does not have the sedtrails_data as input. It is the latest test that I added for the plugins. Should I modify the `soulsby.py` script? It is a placeholder right now anyways.

I understand now. Natasha's code is still on review. Better not to touch the soulsby.py file for now. Let's keep it as is

aysunrhn:
- I modified the plugin parameter test to check that sedtrails_data is in the correct order.
- Updated the print message to state which plugin file is failing the test:

PR #267, opened 2025-08-13
Particle seeding

manuGil:
dThis PR contains changes to the particle seeding. 

The following particle strategies have been implemented:

- Point
- Random
- Grid
- Transect

Adds a ParticleFactory to create any number of particles.
This also contains changes to the config schema. Changes were necessary to modify the options that each seeding strategy provides. 

## Reasoning and Usage

In the configuration file the particle for a simulation are defined using populations. A population is a collection of particles of the save type and characteristics, which have the same tracer methods and seeding strategy:

```yml
particles:
  population:
    name: my_population
    particle_type: sand
    characteristics:
      grain_size: 0.01 
      density: 2650.0  
    tracer_methods:
      vanwesten:
        alpha: 0.2
        beta: 0.3
    seeding:
      release_start: 2025-06-18 13:00:00
      quantity: 1
      strategy: 
        point:
          locations: 
            - "40000,17000"
```

> Note that a simulation can have more than one population. And that if we want to include particles of different types or characteristics, or seed particles using different strategies, we shall create different populations.

The code in this PR follows this reasoning:
- The seeding tool generates n-number of particles for a population using the seeding strategy defined in the configuration for each population.  
- The number of particle per location are  defined by `quantity` and the number of seeding locations depend on the strategy used. e.g., above only one seed location is specified. The number of seeding locations for some of the strategies are computed at runtime (e.g., grid).
- The total number of particle in a population is the product of `quantity * seeding locations` 
- When seeding happens a list with the total number of particles in a population is created. If more than one population is specified in the configuration file, than particle creation should be repeated for each one (I will work on automating this). The creation of particle is manage by a `ParticleFactory`, which takes the configuration for a population and generates the particles, assigns the initial x,y coordinates, release_time (release_start) and quantity. 

The `ParticleFactory` can be used as follows:

```python
    # Example usage
config_point = SeedingConfig(
    {
        'population': {
            'seeding': {
                'strategy': {'point': {'locations': [
                                                '1.0,2.0', 
                                                '3.0,4.0'
                                               ]
                                 }},
                'quantity': 1,
                'release_start': '2025-06-18 13:00:00'
            }
        }
    }
)

# create 
particles = ParticleFactory.create_particles(config_point)
```
Which returns something like (1*2 number of particles):

```python
[
Sand(id=1, _x=1.0, _y=2.0, _release_time='2025-06-18 13:00:00', _is_mobile=True, name='', trace={}, physical_properties=PhysicalProperties(density=2650.0, diameter=0.0002)), 
Sand(id=2, _x=3.0, _y=4.0, _release_time='2025-06-18 13:00:00', _is_mobile=True, name='', trace={}, physical_properties=PhysicalProperties(density=2650.0, diameter=0.0002))
]
```
Particle `id` is automatically assigned.

### DIFF
--- a/src/sedtrails/particle_tracer/particle_seeder.py
+++ b/src/sedtrails/particle_tracer/particle_seeder.py
@@ -65,7 +65,7 @@ class SeedingStrategy(ABC):
     """
 
     @abstractmethod
-    def seed(self, config: SeedingConfig) -> List[Tuple[int, float, float]]:
+    def seed(self, config: SeedingConfig, *args) -> List[Tuple[int, float, float]]:
         """
         Asociates quantity of particles to a seeding locations for a given strategy.
 
@@ -73,6 +73,8 @@ class SeedingStrategy(ABC):
         ----------
         config : SeedingConfig
             Configuration object containing the seeding parameters.
+        *args :
+            Additional positional arguments that may be used by specific strategies.
 
         Returns
         -------
@@ -90,7 +92,7 @@ class PointStrategy(SeedingStrategy):
     Seeding strategy to release particles at specific locations (x,y).
     """
 
-    def seed(self, config: SeedingConfig) -> list[Tuple[int, float, float]]:
+    def seed(self, config: SeedingConfig, *args) -> list[Tuple[int, float, float]]:
         locations = find_value(config.population_config, 'population.seeding.strategy.point.locations', [])
         if not locations:
             raise MissingConfigurationParameter('"locations" must be provided for PointStrategy.')
@@ -114,7 +116,7 @@ class RandomStrategy(SeedingStrategy):
     Seeding strategy to release particles at at random locations (x,y) within an area constraint by a bounding box 'xmin,ymin xmax,ymax'.
     """
 
-    def seed(self, config: SeedingConfig) -> list[Tuple[int, float, float]]:
+    def seed(self, config: SeedingConfig, *args) -> list[Tuple[int, float, float]]:
         bbox = find_value(config.population_config, 'population.seeding.strategy.random.bbox', {})
         print(bbox)
         if not bbox:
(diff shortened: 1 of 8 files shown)


src/sedtrails/particle_tracer/particle_seeder.py
```diff
@@ -54,354 +65,262 @@ class SeedingStrategy(ABC):
...
+        quantity = int(config.quantity)
+        seed_locations = []
+        dx = find_value(grid, 'separation.dx')
+        dy = find_value(grid, 'separation.dy')
+        xvals = []
+        yvals = []
+        x = bbox['xmin']
+        while x <= bbox['xmax']:
+            if x > bbox['xmax']:
+                break
```
aysunrhn, line 182: Is it necessary to have the `x > bbox['xmax']` if statement? Doesn't the while program exit the while loop as soon as this condition is satisfied already?

src/sedtrails/particle_tracer/particle_seeder.py
```diff
@@ -54,354 +65,262 @@ class SeedingStrategy(ABC):
...
-            # TODO: find if it is needed to round the position to a specific number of decimal places
-            particle = Passive(id=i, _x=initial_position[0], _y=initial_position[1], _release_time=release_times[i])
-            particles.append(particle)
-
-        return particles
+        _bbox = bbox.replace(',', ' ').split()  # separates values with whitespaces. Order is xmin, ymin, xmax, ymax
+        for _ in range(quantity):
+            x = random.uniform(float(_bbox[0]), float(_bbox[2]))
+            y = random.uniform(float(_bbox[1]), float(_bbox[3]))
+            seed_locations.append((quantity, x, y))
```
aysunrhn, line 150: Do you think it is a better approach to have a fixed seed for the random number generations at the beginning at least to make sure everything is deterministic? I don't know how much the seeding randomness effects the overall functioning, though maybe it's negligable.

tests/particle_tracer/test_particle_seeder.py
```diff
@@ -1,70 +1,426 @@
 """
```
aysunrhn, line 1: Are there any edge cases for the seeding strategies that we might want to test? Things that are physically impossible.

src/sedtrails/particle_tracer/particle_seeder.py
manuGil: You are rite about the while loop. I added that extra check to guarantee no `x` is created outside the bounding box. I could remove it, but I feel that gives me some peace of mind 🤓

src/sedtrails/particle_tracer/particle_seeder.py
manuGil: Your comment is on spot! However, I think the answer lies onto the approach to randomness a coastal engineer usually follows. Maybe @sgpearson17  can tell us more about that.  For the goal of this feature, I think the implementation of randomness is not crucial.

tests/particle_tracer/test_particle_seeder.py
manuGil: One that comes to my mind:
- A particle cannot have a location that fall outside the input-data's spatial domain (this is one of the reasons we need the dimensions of the input data).

manuGil: @sgpearson17 I think the current seeding tool (this code) only works for the positive side of coordinate system, but not the negative side. Do you expect to work with input data that may contain positive and negative coordinate values?

manuGil:
@Hassan-Shafiei I added an brief description of how seeing works on the top of this discussion.

src/sedtrails/particle_tracer/particle_seeder.py
sgpearson17: The random function was mainly used in the past to avoid problems in visualization associated with a regularly-spaced grid of particles, and I wasn't thinking much about reproducibility/determinism. But in many of our coastal models (e.g. wave generation in XBeach), the default is to always use a random seed, but that there is usually an input parameter/flag where a user can specify a fixed seed (e.g., randomSeed=1) so that you can check isolate that randomness when trying to debug your model. This would be a helpful feature to implement, if you can? Happy to discuss further!

tests/particle_tracer/test_particle_seeder.py
sgpearson17: Good questions! Indeed a particle cannot go outside the input model's spatial domain (nothing to move it there), but I suppose there may be some coordinate systems with negative coordinate values. I did a quick check and the two most likely ones we will use (RD New Amersfoort in Netherlands and UTM everywhere else) are both always positive. Lat/Lon can have negative values but we need users to provide coordinates in metres for calculation purposes. I don't anticipate doing a lot of ocean-scale modelling in spherical coordinates (we are primarily concerned with sediment transport on smaller scales) so implementing such a feature is not a priority. 

Concluding: 
1. Particles can't go outside the input model doimain
2. Let's assume "always positive" but make a note of that in the docstrings/comments.

src/sedtrails/particle_tracer/particle_seeder.py
manuGil: I will add add `seed` parameter under the random strategy, so that we can have a reproducible outcome

manuGil:
@aysunrhn  Some of the check are failing, but that's because we have to clean code related with other components.

aysunrhn:
> @aysunrhn Some of the check are failing, but that's because we have to clean code related with other components.

I will merge this PR then.

aysunrhn (approved):

PR #272, opened 2025-08-13
Polish NetCDFReader class and add tests

### LINKED ISSUE #234: Link simulation through input reader
Link simulation output file (sedtrails as netCDF) to netCDF in the Data Manager.
A use case for this is the need to read the results of a simulation to perform a connectivity analysis.

aysunrhn:
Updated the NetCDFReader class in `netcdf_reader.py` to be handle file loading errors more cleanly and added accompanying tests.

Closes #234 

# Relevant files
- `src/sedtrails/data_manager/netcdf_reader.py`
    - Added multiple file validations (correct file extension, file exists, and if it can be opened with xu)
    - Switched to `xu.open_dataset` because NetCDFReader will handle single file instead of multiples
    - Updated docstrings

- `tests/data_manager/test_netcdf_reader.py`
    - Tests cover missing file, bad extension, readable file and supported extensions (".nc", ".nc4", ".netcdf")

### DIFF
--- a/src/sedtrails/data_manager/netcdf_reader.py
+++ b/src/sedtrails/data_manager/netcdf_reader.py
@@ -12,23 +12,68 @@ class NetCDFReader:
     """
     A class for reading NetCDF files produced by the SedTrails Particle Tracer System.
 
+    This class validates and loads NetCDF files using xugrid's lazy loading functionality.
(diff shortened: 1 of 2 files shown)


manuGil (approved):
This looks good, and it can be merged

PR #274, opened 2025-08-13
Implement SedtrailsMetadata class and integrate with SedtrailsData

### LINKED ISSUE #273: Implement Sedtrails metadata class
- We need a Sedtrails metadata class to store metadata (such as the min-max values of the flow field domain).
- This new class should be a required attribute of the existing SedtrailsData class
- The new metadata class should have the min-max values of the flowfield domain as required attributes

aysunrhn:
Added a new `SedtrailsMetadata` class to store flowfield domain bounds and additional metadata (for future), and integrates it as a required field in `SedtrailsData` class. 

Closes #273 

# Relevant files
- `src/sedtrails/transport_converter/sedtrails_metadata.py`
    - SedtrailsMetada dataclass also allows for dynamic attribute addition
    - Required `flowfield_domain`: Contains bounds (x_min, x_max, y_min, y_max)
    - Validates required keys and converts values to float in `__post_init__`
    - Prevents overwriting reserved keys like `flowfield_domain`
    - Example use:
 ```python
# Required domain bounds
metadata = SedtrailsMetadata(
    flowfield_domain={"x_min": 0.0, "x_max": 100.0, "y_min": 0.0, "y_max": 50.0}
)

# Add additional metadata dynamically
metadata.add("model_type", "delft3d_fm")
metadata.update({"timestep": 600, "units": "meters"})

# Access domain bounds
print(metadata.x_min)  # 0.0
```

- `src/sedtrails/transport_converter/sedtrails_data.py`
    - Type checking to ensure metadata is a `SedtrailsMetadata` instance
    - Relies on SedtrailsMetadata's own validation for internal structure

- `src/sedtrails/transport_converter/plugins/format/fm_netcdf.py`
    - Creates `SedtrailsMetadata` with computed domain bounds from coordinate data (`net_xcc` and `net_ycc`)
    - Passes metadata as required field to `SedtrailsData` constructor

### DIFF
--- a/src/sedtrails/transport_converter/plugins/format/fm_netcdf.py
+++ b/src/sedtrails/transport_converter/plugins/format/fm_netcdf.py
@@ -5,6 +5,7 @@ import xarray as xr
 import numpy as np
 from sedtrails.transport_converter.plugins import BaseFormatPlugin
 from sedtrails.transport_converter.sedtrails_data import SedtrailsData
+from sedtrails.transport_converter.sedtrails_metadata import SedtrailsMetadata
 from pathlib import Path
 from typing import Dict, Any, List, Union
 
@@ -120,6 +121,16 @@ class FormatPlugin(BaseFormatPlugin):
             'magnitude': np.zeros_like(depth_avg_velocity_magnitude),
         }
 
+        # Create SedtrailsMetadata object
+        metadata = SedtrailsMetadata(
+            flowfield_domain={
+                'x_min': np.min(mapped_data['x']),
+                'x_max': np.max(mapped_data['x']),
+                'y_min': np.min(mapped_data['y']),
+                'y_max': np.max(mapped_data['y']),
+            }
+        )
+
         # Create SedtrailsData object
         sedtrails_data = SedtrailsData(
             times=seconds_since_ref,
@@ -136,6 +147,7 @@ class FormatPlugin(BaseFormatPlugin):
             max_bed_shear_stress=mapped_data['max_bed_shear_stress'],
             sediment_concentration=mapped_data['sediment_concentration'],
             nonlinear_wave_velocity=nonlinear_wave_velocity,
+            metadata=metadata,
         )
 
         return sedtrails_data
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/sedtrails/transport_converter/plugins/format/fm_netcdf.py [hunk 1]
```diff
@@ -120,6 +121,16 @@ def convert(self) -> SedtrailsData:
             'magnitude': np.zeros_like(depth_avg_velocity_magnitude),
         }
 
+        # Create SedtrailsMetadata object
+        metadata = SedtrailsMetadata(
+            flowfield_domain={
+                'x_min': np.min(mapped_data['x']),
```
aysunrhn, line 127: Check if the min() method is calculating the min values on the fly.

aysunrhn: > Check if the min() method is calculating the min values on the fly.

@manuGil 
Reading the xarray documentation, it appears that calling `mapped_data['x'].min()` also calculates the minimum, it does not access a pre-computed value but it might be faster because it doesn't convert to np.array. But that means it returns an DataArray object with metadata (which coordinate it calculated from). 

It seems to me like keeping the np.array conversion might be a better idea since we don't need this metadata, and only a scalar value.

manuGil: Aysun, has this been checked?

aysunrhn:
@manuGil  I checked the `mapped_data['x'].min()` function last week but for some reason my comment is left as "pending", so here is what I found out last week:

Reading the xarray documentation, it appears that calling mapped_data['x'].min() also calculates the minimum, it does not access a pre-computed value but it might be faster because it doesn't convert to np.array. But that means it returns an DataArray object with metadata (which coordinate it calculated from).

It seems to me like keeping the np.array conversion might be a better idea since we don't need this metadata, and only a scalar value.

src/sedtrails/transport_converter/plugins/format/fm_netcdf.py
manuGil: Alright, let's move on with this then.

manuGil (approved):
Please merge this.

PR #275, opened 2025-08-13
Fix/porosity

manuGil:
This PR add additional parameters to the Physic section of the json schema. And fixes an issue that resulted from injecting default values after config file validation.

### DIFF
--- a/src/sedtrails/config/main.schema.json
+++ b/src/sedtrails/config/main.schema.json
@@ -340,10 +340,10 @@
               "default": 2.5e-4,
               "description": "grain diameter of sediment [m]"
             },
-            "morfology_factor": {
+            "morphology_factor": {
               "type": "number",
               "default": 1.0,
-              "description": "morphology factor for sediment transport (1.0 = no effect)"
+              "description": "morphological acceleration factor (Roelvink, 2006) (1.0 = no effect)"
             },
             "porosity": {
               "type": "number",
--- a/src/sedtrails/config/main.schema.json
+++ b/src/sedtrails/config/main.schema.json
@@ -333,7 +333,23 @@
       "properties": {
         "constants": {
           "type": "object",
+          "additionalProperties": false,
           "properties": {
+            "grain_diameter": {
+              "type": "number",
+              "default": 2.5e-4,
+              "description": "grain diameter of sediment [m]"
+            },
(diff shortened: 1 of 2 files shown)


src/sedtrails/config/main.schema.json
```diff
@@ -333,7 +333,23 @@
...
         "constants": {
           "type": "object",
+          "additionalProperties": false,
           "properties": {
+            "grain_diameter": {
+              "type": "number",
+              "default": 2.5e-4,
+              "description": "grain diameter of sediment [m]"
+            },
+            "morfology_factor": {
```
sgpearson17: typo: morfology --> morphology

@bartvanwesten : is this in here from your code? do you think we should directly specify this here? We don't have any kind of real physics-based sedtrails acceleration factor yet, so I wonder if it's better to take this out so that people don't get any ideas. Is this used anywhere yet? I would be in favour of removing it. @jreyns - thoughts?

src/sedtrails/config/main.schema.json
```diff
@@ -333,7 +333,23 @@
...
+              "type": "number",
+              "default": 1.0,
+              "description": "morphology factor for sediment transport (1.0 = no effect)"
```
sgpearson17: this should be "morphological acceleration factor (Roelvink, 2006) (1.0 = no effect)"

sgpearson17 (changes requested):
Seems good to me, I just have a few questions about why the morfac parameter is in there and where it gets used.

src/sedtrails/config/main.schema.json
manuGil: Thanks for the comments @sgpearson17 I corrected the description and the typo. I will wait for Bart or Johan's reactions before I complete the changes.

sgpearson17:
I checked with Bart just now and apparently he is actually using that parameter in a pre-processing step of the dfm model results before bringing it into sedtrails, so maybe let's indeed include it for now. This PR is good to go, then.

sgpearson17 (approved):
I checked with Bart just now and apparently he is actually using that parameter in a pre-processing step of the dfm model results before bringing it into sedtrails, so maybe let's indeed include it for now. This PR is good to go, then.

PR #281, opened 2025-08-21
started writing seeding documentation although things are changing

Hassan-Shafiei:
This is just the start of preparing the documentation. It still requires major revisions, especially considering that the seeder scripts are under development.


manuGil (approved):
Thank you @Hassan-Shafiei You can merge this

PR #282, opened 2025-08-21
final revisions of the diagrams

Hassan-Shafiei:
I corrected some typos, but maybe the diagrams might still need some revisions following reaching the first milestone? 
Also, the diagrams of the connectivity app seems incomplete.

### DIFF
--- a/diagrams/components-cache-app.puml
+++ b/diagrams/components-cache-app.puml
@@ -6,7 +6,7 @@
 title SedTRAILS: Component Diagram for Simulation Cache and Recovery Application
 scale 1.5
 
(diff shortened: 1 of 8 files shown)


manuGil:
@Hassan-Shafiei Thanks for check the diagrams. You're right, the diagrams about the connectivity component are incomplete. The will remain like that during our support. When we we working on them, it was decided that we will get back to them if time allowed it. Realistically, that won't happened during the next month. Nevertheless, the existing diagrams still portrait the general structure of the component and how we ambition to integrate it into the rest.

manuGil (approved):
Thanks for checking this thoroughly.

PR #302, opened 2025-08-27
Fix timestep deviation check

### LINKED ISSUE #301: Check why timestep troubles detected "Found 10 timesteps deviating from median (600s)"????  Is it looping?

aysunrhn:
To check if any timestep is deviating from the median, we were comparing each time point against the median. Instead, we should be comparing the time differences against the median value.

Closes #301

### DIFF
--- a/src/sedtrails/transport_converter/sedtrails_data.py
+++ b/src/sedtrails/transport_converter/sedtrails_data.py
@@ -106,6 +106,7 @@ class SedtrailsData:
             timestep = None
         else:
             # Calculate median timestep, this helps ignore the weird startup timesteps
+            all_timesteps = np.diff(self.times)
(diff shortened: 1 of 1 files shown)


src/sedtrails/transport_converter/sedtrails_data.py
```diff
@@ -114,7 +115,7 @@ def _calculate_timestep(self):
 
             # Check if we have timesteps deviating from the median
             tolerance = 1e-6
-            deviations = np.abs(self.times - timestep)
+            deviations = np.abs(all_timesteps - timestep)
```
sgpearson17, line 118: yup, makes sense!

sgpearson17 (approved):
Yup, that makes sense. Good catch!

PR #303, opened 2025-08-27
Add simulation dashboard for real-time visualization

bartvanwesten:
Introduces a new SimulationDashboard class and integration in simulation.py for real-time monitoring and visualization of particle simulations. The dashboard provides spatial and temporal plots, including flow field, bathymetry, particle trajectories, and time series statistics. Dashboard initialization and updates are now triggered during simulation steps if enabled.


src/sedtrails/simulation.py
```diff
@@ -328,8 +329,49 @@ def run(self):
...
+                particle_data = {
+                    'x': first_population.particles['x'],
+                    'y': first_population.particles['y'],
+                    'burial_depth': first_population.particles['burial_depth'],
+                    'mixing_depth': first_population.particles['mixing_depth']
+                }
+                
+                # Get simulation timing
+                # plot_interval_str = self._controller.get('output.plot_interval', '1H')
+                plot_interval_seconds = 3600 # self._parse_duration(plot_interval_str)
```
sgpearson17, line 358: we should move this (and other things? ``enable_dashboard = True?``) to the .yaml file

sgpearson17 (commented):
**First impressions:**
It works! From a functional perspective it is useful and the default fields are at least for now good/interesting choices. We can always customize further down the road.
image

It seems to run a bit slowly so I tried to compare running the same simulation with and without the dashboard running:
With dashboard and plot updating @ 600s: ``09:46``
Without dashboard: ``09:31``
So it runs a bit faster without it but it doesn't seem to be a fatal bottleneck.

My only real comments are from an integration perspective:
1. I couldn't just set ``enable_dashboard = False`` and run it- I needed to comment out the whole block of code. This should probably go in the yaml file, but we should also make it possible to turn it on/off easily. 
2. How well does this integrate with the rest of the code? I guess that is something for @manuGil to look at when he is checking ``simulation.py``?

src/sedtrails/simulation.py
manuGil: I will make changes to the visualisation schema with that option

manuGil (approved):
Thanks Bart. I will merge it, and if necessary integrate the dashboard further into the simulation workflow

PR #305, opened 2025-08-27
FIx simulation loop to avoid recreating sedtrails_data unnecessarily

### LINKED ISSUE #304: Sedtrails_data is initialized multiple times unnecessarily
In `simulation.py` there initial if condition forces reloading and reconverting `sedtrails_data` object from scratch until the current time falls within the simulation time. 

https://github.com/aysunrhn/sedtrails/blob/b51453d024e98fc9ef313c203842dcd8c39a58fb/src/sedtrails/simulation.py#L266-L273

```python
        while not timer.stop:
            # Check if current time is within loaded SedTRAILS data
            current_time_seconds = timer.current
            if (
                sedtrails_data is None
                or current_time_seconds < sedtrails_data.times[0]
                or current_time_seconds > sedtrails_data.times[-2]
            ):
```

aysunrhn:
A quick fix to the simulation time loop so we avoid recreating the `sedtrails_data` and converting multiple times from scratch before the simulation has started. 

Closes #304 

https://github.com/aysunrhn/sedtrails/blob/0d2c0a480db73c855eab5ec78966098f41010621/src/sedtrails/simulation.py#L266-L275

```python
            # Check if current time is within loaded SedTRAILS data
            current_time_seconds = timer.current
            if (
                sedtrails_data is None
                or current_time_seconds > sedtrails_data.times[-2]
            ):
                # Avoid recreating SedTRAILS data if current time is before the first time step
                if current_time_seconds < sedtrails_data.times[0]:
                    timer.advance()
                    continue
```

### DIFF
--- a/src/sedtrails/simulation.py
+++ b/src/sedtrails/simulation.py
@@ -272,7 +272,7 @@ class Simulation:
                 # Avoid recreating SedTRAILS data if current time is before the first time step
                 if current_time_seconds < sedtrails_data.times[0]:
                     timer.advance()
-                    pass
+                    continue
                 # Convert to SedTRAILS format
                 sedtrails_data = self.format_converter.convert_to_sedtrails(
                     current_time=current_time_seconds, reading_interval=simulation_time.read_input_timestep.seconds
--- a/src/sedtrails/simulation.py
+++ b/src/sedtrails/simulation.py
@@ -267,9 +267,12 @@ class Simulation:
             current_time_seconds = timer.current
             if (
                 sedtrails_data is None
-                or current_time_seconds < sedtrails_data.times[0]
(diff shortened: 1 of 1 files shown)


manuGil (approved):
Thanks. Let's keep an eye on other improvements like this.

manuGil:
I noticed that this solution wasn't completed. The code failed on the second if, because sedtrails_data has no times attribute before initialization. I already added a fix

aysunrhn:
> I noticed that this solution wasn't completed. The code failed on the second if, because sedtrails_data has no times attribute before initialization. I already added a fix

I overlooked that bit, my apologies! What was the fix? Is it in a new PR?

PR #309, opened 2025-08-27
Visualization Dashboard

manuGil:
Improvements to the integration of the visualization dashboard with the simulation class. 
The code can could be made cleaner, and implement asynchronous operations to update the dashboard while the simulation runs (using `asyncio`).

### DIFF
--- a/examples/config.example.yaml
+++ b/examples/config.example.yaml
@@ -56,8 +56,10 @@ particles:
             bbox: "39400,16800 40600,17800"
             seed: 42
             nlocations: 5
-            
 output:
   store_tracks: true
   save_interval: 1H
   plot_interval: 1D
+visualization:
+  dashboard:
+    enable: true
--- a/examples/config.example.yaml
+++ b/examples/config.example.yaml
@@ -7,7 +7,7 @@ folder_settings:
   input_data: ./sample-data/inlet_sedtrails.nc
   output_dir: ./results
 time:
-  start: 2011-08-01 00:00:00
+  start:  2016-09-21 19:20:00
   timestep: 60S
   read_input_timestep: 1D
   duration: 1D
--- a/src/sedtrails/config/visualization.schema.json
+++ b/src/sedtrails/config/visualization.schema.json
@@ -183,6 +183,17 @@
                     "description": "settings for validation plots"
                 }
             }
+        },
+        "dashboard": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "enable": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Enable real-time dashboard? true/false"
+                }
+            }
         }
     }
 }
--- a/src/sedtrails/pathway_visualizer/__init__.py
+++ b/src/sedtrails/pathway_visualizer/__init__.py
@@ -1,1 +1,5 @@
 """A directory for the Pathway Visualizer"""
+
+from .simulation_dashboard import SimulationDashboard
+
+__all__ = ['SimulationDashboard']
--- a/src/sedtrails/pathway_visualizer/simulation_dashboard.py
+++ b/src/sedtrails/pathway_visualizer/simulation_dashboard.py
@@ -12,7 +12,6 @@ from collections import defaultdict
 from pathlib import Path
 import datetime
 from typing import Dict, Tuple
-import asyncio
 
 
 class SimulationDashboard:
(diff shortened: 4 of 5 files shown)


manuGil:
I continued testing the dashboard this morning because I found some quirks while working on McOS. I made several more changes to improve when the dashboard window is opened and closed. Then, I noticed that my plots don't look like the one by @sgpearson17 in #303. Specially the flow-field. I always get a green square .  I went back to the original pull request by @bartvanwesten  #303 to check if the changes on the display were caused by my changes to the code. In both cases, the display shows something very similar (i.e. not what Stuart posted). 
Could you both confirm that I did not break the dashboard? ;-) Could the changes in the display be explain by the input dataset (I am still using the old inlet file), or the values in the config file?

Below what I see:

## Original Bart PR:

Screenshot 2025-08-28 at 10 25 09

## After my Changes (This PR)
I assume line-plots will be different because I am using a random seeding method.
 
 
Screenshot 2025-08-28 at 10 37 46

sgpearson17:
Hi Manuel, I was actually just looking at this pull request now- I can't turn the dashboard back on- where should I enable it in the yaml? I noticed you had turned it off by default now, but when I tried ``dashboard: true`` in the main config file it wasn't happy with me. I noticed that it runs about 10x faster now than the one I tested yesterday (for no-dashboard cases).

The only thing I can think of is that maybe the plot interval is low? the simulation in your config example is 1 day long, and the plot interval is also 1D (``output:  plot_interval: 1D``), so maybe that's why you just see flat lines?

sgpearson17:
Also, indeed maybe the input file makes a difference. I re-ran it with Bart's old branch (https://github.com/sedtrails/sedtrails/tree/visualization_dashboard) and the old input file (``inlet_sedtrails.nc``)

**Old Branch, Old Input:**
image

**Old Branch, New Input:**
image

It is much much slower with the new input! Since the old input contains an "empty" timestep at the beginning, I wonder if fewer particles are considered active (see green flow field in old-old example above).  If more particles are active I imagine it would take longer.

Also also, I see just now that your latest commit adds the dashboard config to the example yaml, so that should be that covered. Will try that now.

sgpearson17:
Changing the start time to ``2016-09-21 19:20:00`` made the old branch-new input version run MUCH faster (20s):

**Old Branch - New Input:**
image

The old branch with old input now shows very similar behaviour with a similar computation time to the old-new version above (differences reasonable given random seeding):
**Old Branch - OLD Input**
image

So I think the start time is the source of our problem. I think the previous start time of ``2011-08-01 00:00:00`` was probably copied over from Bart's Sand Engine model that he was using to do tests on his local machine. Will try the new PR branch version now.

sgpearson17:
Now here is the new branch (this PR):

New branch, old file:
image

New branch, new file:
image

Seems close enough to me (explainable with randomization). So I think the start date was indeed the problem.

sgpearson17 (approved):
Dashboard works as intended (see earlier comments re: tests). Approved but I think we should make this change to the default setting in yaml file:
```
time:
  start: 2016-09-21 19:20:00
```
Feel free to merge.

manuGil:
Hello @sgpearson17  Thanks for your solving this. I was busy with other things. But I am can see you're on your way to be a 10x software tester 🙌  . 
I will merge the current changes. However, I want to make the following remarks regarding the speed, and the effect of the time value on the dashboard. 

1. In the current implementation, with the dashboard enabled, updating the dashboard is a 'blocking operation'. The simulation must wait until the dashboard is updated before it continues with the next time step. We could improve that by implementing asynchronous call to update the dashboard. Yesterday, I took a look to how to do that, but realized it wasn't a quick fix as I originally thought. We should come back to this later. 

2. We should consider if we can implement the dashboard in such a way that it doesn't behave 'strange' when values for `time.start` are too different than the ones in the input data. I particularly wonder if we can make changes and decision on the code so that, the dashboard doesn't depend on the value of `time.start` directly but on the effect that `time.start` have on the data and values that we want to visualize. In other words, maybe we can think on a way in which the dashboard doesn't depend (use) that value directly.

sgpearson17:
Sounds good! I added an issue for the asynchronous call. #310

manuGil:
> Sounds good! I added an issue for the asynchronous call. #310

Thanks. I noticed a small issue on the display that it is somehow related to this point. The progress bar on the display never shows 100%. That has to do partly due the dependency between the main loop and updating the display. I expect a solution that this will be evident when we decouple the dashboard from the main simulation loop.

Issue #321, opened 2025-09-10
Multiple result folders are created even at first simulation run


aysunrhn:
Multiple result folders are created even when we run the simulation for the first time, the logger and data manager module are conflicting as they both want to create results folder without overwriting an existing one.

aysunrhn:
Closed in #323

PR #322, opened 2025-09-10
Fix logger and output file handling

### LINKED ISSUE #288: Clean up the logger class to have flexible states
The predefined logging states are no longer necessary, we need to remove them.

aysunrhn:
Added a centralized logging module which is initialized in simulation.py and the logger output is the same as the simulation results output.

Closes #288 #321

- Logger is initialized in``simulation.py` after configuration is loaded.
- Provides a single, global logger (“sedtrails”) usable from any module as:
```python
import logging
logger = logging.getLogger(__name__)
logger.info("Insert your log message here")
```
- Adds a global exception hook so unhandled exceptions are written to the log.
- The structured log messages we use during simulation run are still there and can be accessed via `logger.log_simulation_state`
- The logger initialization comes after DataManger is initialized, which internally initializes NetCDFWriter and creates a new results folder with the timestamp appended if it already exists. This way, the logger is initialized in the correct results folder (which is defined by the NetCDFWriter internally).

### DIFF
--- a/src/sedtrails/configuration_interface/configuration_controller.py
+++ b/src/sedtrails/configuration_interface/configuration_controller.py
@@ -9,11 +9,9 @@ and provides configurations to other components.
 from abc import ABC, abstractmethod
 import sys
 import os
-import logging
 
 from sedtrails.configuration_interface.validator import YAMLConfigValidator
 from sedtrails.configuration_interface.find import find_value
-from sedtrails.logger.logger import setup_logging
 from typing import Dict, Any, Optional
 
 class Controller(ABC):
@@ -100,10 +98,6 @@ class ConfigurationController(Controller):
             validator = YAMLConfigValidator()
             self.config_data = validator.validate_yaml(config_file)
 
-            self.init_logging_from_config()
-            logger = logging.getLogger(__name__)
-            logger.info("Configuration loaded")
-
         return None
 
     def get_config(self) -> dict:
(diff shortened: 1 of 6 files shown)


manuGil:
@aysunrhn Please fix the Ruff errors

manuGil:
> @aysunrhn Please fix the Ruff errors

I will solve them myself this time. I have to update my fork to avoid more conflicts. Please always run `ruff check ...` before push code in the future.

manuGil (approved):

manuGil:
I noticed that the issue with overwriting the log file still persists.

aysunrhn:
> I noticed that the issue with overwriting the log file still persists.

When you run a new simulation (while you still have the "results" folder) do you get a new folder? For me it was creating a new folder and a new log file at each simulation, I wonder if the changes in the NetCDFWriter affected anything. I will look into this now.

manuGil:
If the changes to the netcf writer is causing that behaviour,  our solutions aren't robust enough, or the code is not well organised.  Let me know what you find.

aysunrhn:
> If the changes to the netcf writer is causing that behaviour, our solutions aren't robust enough, or the code is not well organised. Let me know what you find.

You are correct. After merging the PRs I noticed the NetCDFWriter no longer creates a new directory if it already exists, and it overwrites the results. But the log file was simply being appended. I will change the behavior now that the log file is also overwritten.

PR #324, opened 2025-09-11
Add NetCDF inspector tool for particle trajectory analysis

sgpearson17:
- Standalone script to inspect SedTrails NetCDF output files
- Shows comprehensive metadata and file structure
- Creates 4-panel visualization:
  * Individual particle trajectories
  * Distance from initial position over time
  * Population-colored trajectories with legend
  * Population statistics with mean and std deviation bands
- Command-line interface with options to save plots
- Supports multiple populations and flow fields

run with:
``python src\sedtrails\pathway_visualizer\nc_inspector.py --save-plot``

### DIFF
--- a/src/sedtrails/pathway_visualizer/nc_inspector.py
+++ b/src/sedtrails/pathway_visualizer/nc_inspector.py
@@ -1,0 +1,395 @@
+#!/usr/bin/env python3
+"""
+NetCDF Inspector for SedTrails Results
+
+A standalone script to inspect and visualize SedTrails NetCDF output files.
+Displays metadata and plots particle trajectories.
+
+Usage:
+    python nc_inspector.py [path_to_netcdf_file]
+    
+If no file is provided, defaults to 'results/sedtrails_results.nc'
+"""
+
+import argparse
+import sys
+from pathlib import Path
+import numpy as np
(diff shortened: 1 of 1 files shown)


aysunrhn:
I have tested the code as well and it seems to run with no issues. Except there are a few warnings, you might want to look at:

image

manuGil:
> I have tested the code as well and it seems to run with no issues. Except there are a few warnings, you might want to look at:
> image

The message is about a future deprecation of the way to access data arrays. We can safe it merge it, because we are still considering if the code should be integrate into the tool.

src/sedtrails/pathway_visualizer/nc_inspector.py
```diff
@@ -0,0 +1,395 @@
...
+    print(f"Inspecting NetCDF file: {nc_file}")
+    
+    try:
+        # Open the NetCDF file
+        with xr.open_dataset(nc_file) as ds:
+            # Print metadata
+            print_metadata(ds)
+            
+            # Show population information
+            inspect_populations(ds)
```
manuGil, line 381: I think we can integrate this with the CLI as a data analysis tool. The analysis is not very advanced in this case but that's a good start. I will do this

manuGil (approved):

PR #326, opened 2025-09-11
Overwrite log file instead of appending it

### LINKED ISSUE #325: Overwrite log file instead of appending
When a new simulation starts in a folder that previously exists, we should overwrite the log file (as we also overwrite the NetCDF output file) instead of appending it.

aysunrhn:
A simple fix to the `logger.py` file so that we overwrite the log file instead of appending to it. 

Closes #325


manuGil (approved):
Thanks.

Issue #336, opened 2025-09-24
Move plugin-specific parts of simulation.py into plugins


sgpearson17:
Move plugin-specific parts of ``simulation.py`` into plugins and out of main loop to preserve modularity? Otherwise we will need to modify ``simulation.py`` a lot anytime we add a new method with different fields (this was the issue we encountered when adding Soulsby, and now there is a mess of if statements).

See ~L350 in ``simulation.py``

sgpearson17:
huh?

Issue #335, opened 2025-09-24
transport_probability_method in Soulsby


sgpearson17:
Right now we just load in transport_probability_method because it is used in vanwesten, but this is not used in Soulsby, so we should clean this up and remove it or enable loading a blank method, since not all other tracer_methods (like passive) will have this stochastic feature

In ``physics_converter.py``, at bottom in ``plugin.add_physics''

sgpearson17:
Partially addressed this in #362 (configured for no_probability transport mode), but now we need to get the stochastic stuff working properly

Issue #334, opened 2025-09-24
implement config controller for multiple populations in a robust way


sgpearson17:
As per this afternoon's discussions with Aysun while debugging Soulsby:

Right now we don't actually use different tracer method per field, we just assume that one is used

L321ish in ``simulation.py``:
```
 # TODO: this loops over populations_config, but only the last one is used. This must be fixed
 # to handle multiple populations with different tracer methods and flow fields
        for population in populations_config:
            tracer_methods = population['tracer_methods']
```

And L122ish in ``simulation.py``:
```
config = PhysicsConfig(
            tracer_method=self._controller.get('physics.tracer_method', 'van_westen'),
```

ALSO: make sure plugin file names in directory match method names (e.g. if you use vanwesten it will look for vanwesten.py, not van_westen.py)

This also shows up in Soulsby: we import ``populations_config`` (as ``tracer_config``?) but don't use all the data

sgpearson17:
johan fixed #391

PR #337, opened 2025-09-24
Fix soulsby

### LINKED ISSUE #331: Run simulation using Soulsby method
Modifying the configuration file to run Soulsby method passes the YAML file validation, but the `simulation.py` (and maybe other parts of the codebase too?) are strict on running only the Van Westen method. We need to adjust the conditions and get the Soulsby running.

```bash
particles:
  populations:
    - name: populaton_1
      particle_type: sand
      characteristics:
        grain_size: 0.00025 
        density: 2650.0  
      tracer_methods:
        #vanwesten:
        #  flow_field_name: 
        #    - bed_load_velocity
        #    - suspended_velocity
        soulsby:
            background_grain_size: 0.0002
            soulsby_b_e: 1.7e-07
            soulsby_theta_s: 0.1
            soulsby_gamma_e: 0.3
            soulsby_mu_d: 0.9
            soulsby_freedom_factor: 1.0
```

An example from `simulation.py` where "vanwesten" has to be in the `tracer_methods` key to run the simulation.
```python
  # Determine flow field names from configuration
  flow_field_names = []
  for population in populations_config:
      if 'tracer_methods' in population and 'vanwesten' in population['tracer_methods']:
          flow_field_names = population['tracer_methods']['vanwesten']['flow_field_name']
          break  # Use the first population's flow fields for now
```

aysunrhn:
Updated the main simulation file and the physics plugins to run the soulsby method without errors. Although, no particle is moving yet.

Relevant files:
- `src/sedtrails/simulation.py`
- `src/sedtrails/transport_converter/physics_converter.py`
- `src/sedtrails/transport_converter/plugins/physics/vanwesten.py`
- `src/sedtrails/transport_converter/plugins/physics/soulsby.py`

Note that the "vanwesten" plugin was renamed from "van_westen" to "vanwesten" to load the plugin successfully.

Closes #331


sgpearson17:
A few notes for picking this up:

- I started a few issues related to this: #334, #335, #336. These are all things that are maybe suboptimal or detract from the modularity of the code.
- Mostly a Stuart thing for the future: In the code above, the Soulsby method works but the particles did not move. This is likely due to how we implemented a dummy mixing/burial depth and probability fields to mimic the van westen format, even when these fields are not used by soulsby. We also do not use the `soulsby_a` and `soulsby_b` scalar fields now, but these will be used for burial probability.
- Also mostly a Stuart thing for the future: I suggest we start by making sure soulsby works with a concrete bed (i.e. no burial). 

- [ ] **Furthermore, we checked briefly but not extensively whether the Van Westen approach still works after all our modifications. This is an important step before the pull request is merged.**

I will be at a workshop next wednesday but am happy to discuss on other days if needed.

examples/config.example_soulsby.yaml
```diff
@@ -0,0 +1,87 @@
...
+        #  flow_field_name: 
+        #    - bed_load_velocity
+        #    - suspended_velocity
+          soulsby:
+            flow_field_name:
+             - grain_velocity
+             # - soulsby_a
+             # - soulsby_b   
+            background_grain_size: 0.0002
+            soulsby_b_e: 1.7e-07
```
manuGil, line 32: please remove the `soulsby` prefixes. It will be evident by the indentation that they are part of soulsby

src/sedtrails/transport_converter/plugins/physics/soulsby.py
```diff
@@ -12,11 +12,12 @@ class PhysicsPlugin(BasePhysicsPlugin):  # all clases should be called the Physi
     This plugin implements the physics calculations as described in Soulsby et al. (2011).
     """
 
-    def __init__(self, config):
+    def __init__(self, config, tracer_config):
         super().__init__()
         self.config = config
+        self.tracer_config = tracer_config
```
manuGil: `self.config` should contain all configurations for the plugin. What is the reason to have a separated variable with config variables `tracer_config`? If all config parameters come from the yaml file, I discourage to this because it will be difficult to understand the code in the long term. Can you please explain why is this change necessary?

src/sedtrails/simulation.py
```diff
@@ -273,8 +281,11 @@ def run(self):
...
         flow_field_names = []
         for population in populations_config:
-            if 'tracer_methods' in population and 'vanwesten' in population['tracer_methods']:
-                flow_field_names = population['tracer_methods']['vanwesten']['flow_field_name']
+            if 'tracer_methods' in population and ('vanwesten' in population['tracer_methods'] or 'soulsby' in population['tracer_methods']):
+                if 'vanwesten' in population['tracer_methods']:
+                    flow_field_names = population['tracer_methods']['vanwesten']['flow_field_name']
+                elif 'soulsby' in population['tracer_methods']:
+                    flow_field_names = population['tracer_methods']['soulsby']['flow_field_name']
                 break  # Use the first population's flow fields for now
```
manuGil, line 289: I remembered, it was mentioned that  the current solution still picks up  the last tracer method when multiple methods are listed. That should be fixed before sooner than later. @aysunrhn and @sgpearson17  Could you look into a solution for that this week?

manuGil (changes requested):
Please,  address the comments I made. I think the solution to work with Soulsby introduces changes that deviate from good practices. If there a good reason for that, please let me know.

src/sedtrails/simulation.py
sgpearson17: Thanks for looking at this, Manuel. Indeed we came across a number of things that were hard-coded for vanwesten so we had to try and generalize them a bit. When Aysun and I sat together last week to look at this it was sometimes a little unclear how certain data was being passed around to various functions, which I think was how we ended up with `tracer_config` - indeed this should be fixed but we were just trying to get it to work all the way through without crashing. I agree that it's important we get the style consistent with best practices and the rest of the code. Hopefully with fresh eyes we can iterate through it again.

I am at a workshop in Utrecht tomorrow but blocked off time on most of Thursday for sedtrails stuff. @aysunrhn I'm not sure if you have time to look at this tomorrow at all, but if you make a start I can pick up where you leave off on Thursday. Otherwise I can start Thursday.

src/sedtrails/transport_converter/plugins/physics/soulsby.py
aysunrhn: If I understand correctly, `self.config` is a `PhysicsConfig` instance which has the attributes predefined in `src/sedtrails/transport_converter/physics_converter.py`. I thought about adding the parameters from the tracer method itself (for soulsby in this case) to this class, but then that would require additional changes to how it interacts with the vanwesten plugin. 

Instead, I decided to keep these 2 configuration parameters separately (since they are accessed also separately in the `simulation.py` file and initialize the PhysicsConverter with these 2 parameters as separate inputs.

```python
for population in self.populations_config:
    tracer_config = population['tracer_methods']
```

```python
self.physics_converter = PhysicsConverter(self._get_physics_config(), tracer_config)
```

What would you recommend to do instead? Expand the PhysicsConfig class with the tracer methods configurations?

src/sedtrails/simulation.py
aysunrhn: > I remembered, it was mentioned that the current solution still picks up the last tracer method when multiple methods are listed. That should be fixed before sooner than later. @aysunrhn and @sgpearson17 Could you look into a solution for that this week?

In the case where we have multiple populations with different methods, the idea is to have separate instances of PhysicsConverter for each population, is that correct? I can work on this tomorrow if we are going ahead with this solution.

sgpearson17:
As far as I recall/understood, that was indeed the idea, as we might be pulling physics from >=1 different sources (whether from different input models or just using different calculation methods). So i guess using two separate instances works? Unless it is better/faster/more consistent with our existing framework to do it differently?

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

sgpearson17:
@aysunrhn did you end up taking a look at this yesterday? is there anything that I can look at/help with today/tomorrow?

aysunrhn:
> @aysunrhn did you end up taking a look at this yesterday? is there anything that I can look at/help with today/tomorrow?

I looked into the issues @manuGil pointed out and tried to fix the PhysicsConfig class in a more robust way. The tracer method specific parameters are now part of the PhysicsConfig attributes instead of being under "tracer_config". If this solution is good to go, I will have a look at this tomorrow.

sgpearson17 (approved):
Thanks Aysun, I took a quick look through and I think it makes sense to proceed with this. Let me know if you need anything!

manuGil:
@aysunrhn @sgpearson17 What is the status of the issues discuss above? Specially with the PhysicsConverter. Are they solved?

sgpearson17:
I have not looked at this since approving those changes on Friday

aysunrhn:
> @aysunrhn @sgpearson17 What is the status of the issues discuss above? Specially with the PhysicsConverter. Are they solved?

Yes, last week I modified the class and the tracer method specific parameters are now part of the PhysicsConfig attributes instead of being under "tracer_config". If this solution is good to go, I can follow up on the rest and implement use of tracer method per population. I wanted to discuss the solution to make sure at first.

manuGil:
> > @aysunrhn @sgpearson17 What is the status of the issues discuss above? Specially with the PhysicsConverter. Are they solved?
> 
> Yes, last week I modified the class and the tracer method specific parameters are now part of the PhysicsConfig attributes instead of being under "tracer_config". If this solution is good to go, I can follow up on the rest and implement use of tracer method per population. I wanted to discuss the solution to make sure at first.

Go ahead. Maybe we still can complete this by the end of tomorrow. I can take a closer look after our meeting

manuGil (approved):

PR #366, opened 2025-11-08
Fix preallocation bug for nc output

jreyns:


src/sedtrails/simulation_orchestrator/simulation_manager.py
```diff
@@ -391,9 +392,18 @@ def run(self):
                         # Update particle position
                         population.update_position(flow_field=flow_field, current_timestep=timer.current_timestep)
 
-                # Collect data from all populations for this timestep using DataManager
-                self.data_manager.collect_timestep_data(xr_data, populations, timer.step_count, timer.current)
+            # Collect data from all populations for this timestep using DataManager
+            self.data_manager.collect_timestep_data(xr_data, populations, timer.step_count, timer.current)
```
jreyns: @sgpearson17 check this timestep data update that i have taken out out of the population loop. This should be correct like this, or am I missing something?

sgpearson17 (approved):
I looked through the changes and it seems to make sense; I also ran the test and it passed. Hopefully this helps Vassia.

PR #377, opened 2026-01-15
Macdonald 2D Eulerian implementation

the-vdag:
Worked on Mcdonald 2D implementation (the eulerian part, until the calculation of the 2D mean particle velocity - which is around line 190 of the code). Until that point the model works as in the variables take values and things are calculated and there are no syntax bugs.

I left several notes on the code on things to check (e.g. which shear stress to use in the formulas)


sgpearson17:
Continued in new branch/PR #439 after refactoring of dev branch

PR #381, opened 2026-02-10
Add SFINCS support + passive tracer method

jreyns:
- Added SFINCS converter plugin
- Changes and bugfixes for new 'passive' tracer method
- Added tests

### DIFF
--- a/src/sedtrails/simulation_orchestrator/simulation_manager.py
+++ b/src/sedtrails/simulation_orchestrator/simulation_manager.py
@@ -605,10 +605,10 @@ class Simulation:
         for var_name in xr_data.data_vars:
             var = xr_data[var_name]
 
-            if 'time' in var.dims:
+            if 'n_timesteps' in var.dims:
                 # Get the shape and create padding
                 pad_shape = list(var.shape)
-                time_dim_idx = var.dims.index('time')
+                time_dim_idx = var.dims.index('n_timesteps')
                 pad_shape[time_dim_idx] = additional_steps
 
                 # Create NaN-filled array for padding
@@ -618,7 +618,7 @@ class Simulation:
                 # Build coordinates for the padded array
                 pad_coords = {}
                 for dim in var.dims:
-                    if dim == 'time':
+                    if dim == 'n_timesteps':
                         pad_coords[dim] = np.arange(current_size, new_max_timesteps)
                     else:
                         pad_coords[dim] = var.coords[dim]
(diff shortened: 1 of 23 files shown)


the-vdag (approved):
Hello!

I went through the changes in the code, I am not sure how deep the review is expected to be. I looked in more detail through the restructuring of soulsby.py and the way numerical operations are handled (np.divide, np.nan_to_num, np.clip, np.errstate, etc.) -  I’ll take these along in my own code.

Also noted how the tests are set up, which will be helpful for my macdonald tests (still on my to-do list).

Looks good to me, happy to approve the merge request. Once Stuart approves it too, I can run a quick check to see if anything changes for my case. Let me know if you’d like me to look at anything more closely.

Cheers, 
Vassia

sgpearson17:
Hi Vassia/Johan,

Thanks so much! Sorry for my delays, I finally have time to look at this tomorrow.

Stuart

src/sedtrails/simulation_orchestrator/simulation_manager.py
```diff
@@ -385,18 +387,26 @@ def run(self):
...
+            for ip, population in enumerate(populations):
+                for _method in tracer_methods.keys():
                     for flow_field_name in flow_field_names:
                         # Obtain scalar field information
                         # TODO: Consider moving van westen specific fields to the plugin itself
-                        mixing_depth = retriever.get_scalar_field(timer.current, 'mixing_layer_thickness')['magnitude']
                         bed_level = retriever.get_scalar_field(timer.current, 'bed_level')['magnitude']
-                        if self.physics_converter.config.tracer_method == 'vanwesten':
+                        mixing_depth = np.full(bed_level.shape, np.nan)
+                        if tracer_methods[ip] == 'vanwesten':
```
sgpearson17: If you try running vanwesten on the example file this line doesn't actually pick up on it so it skips all the way down to the `else` case. Each population stores tracer_methods as a dictionary, so comparing it directly to vanwesten always fails. If you change the code to extract the actual method name string first, then compare that string, it works for vanwesten and soulsby. I will try to push my changes.

sgpearson17 (commented):
Hi Johan and Vassia,

Sorry it has taken me so long to review and approve this. I realize that you are both out at the moment (and now it's also a long weekend) but I finally had some time to dive deep into the code so I wanted to go for it; you can look at it once you're back.

A few things:
1. The proposed changes *look* good, but I don't have a copy of the `sfincs_map.nc` file, so I cannot fully test it. Can you share a copy of this with me so that I can finish testing? 
2. I noticed you changed a bunch of other stuff in the simulation manager, so I tried testing the code for vanwesten/soulsby/passive using the original dfm nc example file. It wasn't reading in the tracer_methods correctly so I made some changes and pushed them.

@jreyns - when are you back? I don't want to bug you with sedtrails stuff before then, but maybe once you are both back, we should plan an afternoon to all sit in the same room and have a bit of a hackathon to get ourselves on the same page. We had our final class yesterday and now I want to get my head back in the sedtrails game...

Please let me know what you think.

Happy easter!

Stuart

jreyns:
I am closing this one; work is continued in PR ST-388 fix spatial interpolation. That PR has:
- much faster dashboard
- much faster particle position calculator on unstructured large grids
- correct time dim expansion

PR #395, opened 2026-05-13
ST-390 merge multifrac physics into dev

### LINKED ISSUE #390: Allow multi-fraction with different physics
At present, multi-population/multi-method simulation is fragile: physics config is effectively derived from the first population, while tracer methods later get overwritten by loop state. This needs fixing

jreyns:
Merge action for multifraction physics into dev branch with improved interpolation and dashboard

closes #390

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -6,6 +6,7 @@ channels:
 dependencies:
   - python=3.13
   - numpy<2.2,>=1.24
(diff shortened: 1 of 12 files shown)


sgpearson17 (approved):
Approved as discussed so that the merge conflicts don't get out of hand :)

PR #410, opened 2026-06-04
ST-400: XBeach model plugin

### LINKED ISSUE #400: Add XBeach plugin
Add format converter for the Xbeach model

jreyns:
Added a plugin for XBeach. 

For docs:
- this works on the *_mean variables (time-averaged output)
- Variables read from XBeach:
    _thetamean
    zb
    hh
    ue
    ve
    taubx
    tauby
    Susg
    Svsg
    Subg
    Svbg
    ua
    cctot_

Closes #400


sgpearson17:
Exciting! at a user meeting today but will look at this asap.

sgpearson17:
At first glance the changes seem reasonable. Do you already have an example file (yaml + xb nc) to try this out on (or a recommendation)? If not, I can dredge one up from the archives and use that... might be nice to provide an example for users when we release.

jreyns:
> At first glance the changes seem reasonable. Do you already have an example file (yaml + xb nc) to try this out on (or a recommendation)? If not, I can dredge one up from the archives and use that... might be nice to provide an example for users when we release.

I can add an example yaml. My test xb input is 70Mb, so will not post that.

sgpearson17:
Hi Johan, I have been trying all morning but have yet to get this to work. A few notes so far:
- I tried an old example XB file (Santa Rosa [from an old tutorial](https://svn.oss.deltares.nl/repos/xbeach/Courses/DSD_2018/01_basic_morphology/hands-on/SantaRosa/)) but it doesn't read it in--> I think there is some backwards compatibility issue, which is fine, but we should note that.
- from what I can see, it does not work for variable beds (e.g. for running morphodynamically). Maybe we should have the option to read zb instead of just zb_mean? was there a conscious reason not to do that, or was it more about just getting it to run using a file you had on hand?  

I am trying to run my own xb model with the right settings from scratch, but you mentioned that you had an example file of your own? would you still be able to share that with me? even just for my tests? we can try making a lightweight test model later

jreyns:
> Hi Johan, I have been trying all morning but have yet to get this to work. A few notes so far:
> 
>     * I tried an old example XB file (Santa Rosa [from an old tutorial](https://svn.oss.deltares.nl/repos/xbeach/Courses/DSD_2018/01_basic_morphology/hands-on/SantaRosa/)) but it doesn't read it in--> I think there is some backwards compatibility issue, which is fine, but we should note that.
> 
>     * from what I can see, it does not work for variable beds (e.g. for running morphodynamically). Maybe we should have the option to read zb instead of just zb_mean? was there a conscious reason not to do that, or was it more about just getting it to run using a file you had on hand?
> 
> 
> I am trying to run my own xb model with the right settings from scratch, but you mentioned that you had an example file of your own? would you still be able to share that with me? even just for my tests? we can try making a lightweight test model later

Hi @sgpearson17, I wrote this explicitly to work with *_mean variables only. Otherwise, especially in surfbeat, you would get random phase long wave flow fields, which don't mean anything flow- and transportwise (you could argue that zb is the final cumulative result of that, and use that iso of zb_mean). This means that it also works with zb_mean, which is your average bed level over that time avg interval. It is a design choice. If we choose zb iso zb_mean, a requirement is then that tintg==tintm. Don't know if we want that.

jreyns:
The list of variables that are expected now is:

nmeanvar =13
thetamean
zb
hh
ue
ve
taubx
tauby
Susg
Svsg
Subg
Svbg
ua
cctot

One thing to check is the unit of thetamean. The nc metadata says rad, but I am not sure about that.

sgpearson17:
mmmm ok yes now I understand where you're coming from. re-running my demo case with those. will let you know soon.

sgpearson17:
Re: wave angle, indeed the metadata says rad, but the parameter space is 0-180:
image

jreyns:
> Re: wave angle, indeed the metadata says rad, but the parameter space is 0-180: image

Then, we have correct the component split in the wave asymmetry contribution to reflect that. I will commit

sgpearson17:
I tried running the Santa Rosa example with the correct mean outputs and using Soulsby but nothing moves:
image
even though there is definitely stuff happening in the model:
image
When you tested this did you use soulsby/vanwesten/passive?

sgpearson17:
update: same story with van westen: no particles moved

sgpearson17:
I tried your new version on the same santa rosa model and I still don't get any particle motion. did it work for your test model? could you share a copy of that with me so that I can try it on my machine?

jreyns:
> I tried your new version on the same santa rosa model and I still don't get any particle motion. did it work for your test model? could you share a copy of that with me so that I can try it on my machine?

I don't understand:

image

jreyns:
> > I tried your new version on the same santa rosa model and I still don't get any particle motion. did it work for your test model? could you share a copy of that with me so that I can try it on my machine?
> 
> I don't understand:

This is with the Santa Rosa model from the XB skillbed, cut after the first two output timesteps, and with your particle seeding scheme put on the bathy. The surf zone ones clearly move

jreyns:
[xbeach_soulsby_v002.points.txt](https://github.com/user-attachments/files/28831363/xbeach_soulsby_v002.points.txt)
[xbeach_soulsby_v002.yaml](https://github.com/user-attachments/files/28831364/xbeach_soulsby_v002.yaml)

sgpearson17:
Mmm I have a different santa rosa model I think (from early tutorials)- can you link me to the skillbed version?

jreyns:
> Mmm I have a different santa rosa model I think (from early tutorials)- can you link me to the skillbed version?

Your example works fine once you set the dates correct and make the sim longer:

image

If you run from CLI, the model now spits error messages to tell you what is going wrong. In this case, the reference date, release times, and start/stop wrt time data in the XB ouput

jreyns:
[xbeach_soulsby_v002.yaml](https://github.com/user-attachments/files/28831679/xbeach_soulsby_v002.yaml)

sgpearson17:
Houston we have liftoff:
image

sgpearson17 (approved):
Works with santa rosa example now

PR #411, opened 2026-06-04
ST-380: land boundary behaviour

### LINKED ISSUE #270: Address Particle Behaviour at Land-Sea Boundaries
This is a longer-term, lower-priority issue to resolve after the DCC Handover in October, but I wanted to write it down while it was still fresh.

After a discussion with Vassia yesterday about her attempts to use SedTRAILS on an idealized delta, we found that many of her particles were getting stuck on land. This is something I had noticed before in a few of my models but it didn't seem like such a big deal ("hey, the particles are just depositing!"), but as her model has so many channels, the problem was more widespread and harder to ignore. 

I see this problem manifest in the existing sedtrails matlab code in two ways:
1) Particles slow down a lot at boundaries as their velocities are interpolated between U(sea)=U and U(land)=0, in cases where the land point is actually specified in the model and has a (null) velocity.
2) Particle velocities are interpolated *across islands* in cases where islands are simply not included in the D3D grid (e.g. "cut outs" rather than 0 velocities or dry cells). In these cases, I think we just interpolate across the island to the next available grid cell, sometimes leading to weird trajectories. I don't have any examples of this handy but have seen it multiple times.

After looking around to see how other models handle it, it seems both common and non-trivial:

### Parcels:
https://docs.oceanparcels.org/en/latest/examples/documentation_stuck_particles.html
https://docs.oceanparcels.org/en/latest/examples/documentation_unstuck_Agrid.html

### OpenDrift:
https://opendrift.github.io/gallery/example_coastline_options.html
https://opendrift.github.io/_modules/opendrift/models/basemodel.html#OpenDriftSimulation.interact_with_coastline

Based on the above, I think there are workarounds. Something to look at down the road.

jreyns:
**Summary**
- Add configurable inner-boundary polygon masking and boundary-class polygon overrides for `open` and `land` edges.
- Apply boundary-aware particle behavior: open-boundary exits mark particles as left-domain; land-boundary contacts mark particles as beached but eligible for later remobilization.
- Extend FM and SFINCS conversion paths with active mesh filtering, Tekal polygon support, metadata diagnostics, and cached geometry/classification paths.
- Update dashboard rendering to hide left-domain particles, mark beached particles light red, and render particle trails efficiently with `LineCollection`.
- Add documentation, examples, and regression coverage for masking, boundary classification, particle status handling, SFINCS and XBeach support, and dashboard behavior.

Closes #270


sgpearson17:
Hi Johan, I'm looking at the code for this now and it looks like a great contribution. do you have any example polygon files that I can use to test this? otherwise I will need to make them myself

jreyns:
> Hi Johan, I'm looking at the code for this now and it looks like a great contribution. do you have any example polygon files that I can use to test this? otherwise I will need to make them myself

I tested it on the inlet example, but tossed the polygons

sgpearson17:
ok I'll make some little islands for that case then.

sgpearson17:
I played around with this a bit and I think the islands are working:
image
but I tried adding an outer boundary and the particles didn't seem to see it. Will try to look closer tomorrow.

sgpearson17:
Hi Johan,

I tried running it with an assortment of polygons, intending to intercept the particle paths, but as far as I can tell there is no discernible difference in behaviour when I use them:
```
domain:
  pol_file: ./outer_domain.pol
  inner_boundary_pol_files:
    - ./islands_small.pol
  boundary_class_pol_files:
    open:
      - ./tiny_offshore_boundary_edges.pol
    land:
      - ./examples/land_bnd_inlet.pol
 ```

image

`outer_domain.pol`
```
L000001
           5           2 
36200 10000
36200 19500
43800 19500
43800 10000
36200 10000
```

`islands_small.pol`
```
L000001
           5           2
39800 17000
39800 17200
40400 17200
40400 17000
39800 17000
```

tiny_offshore_boundary_edges.pol
```
L000001
           4           2 
39400 16200
39400 17600
40600 17600
40600 16200
```

`land_bnd_inlet.pol`
```
L000001
           5           2
39000 15000
41000 15000
```

Am I running something incorrectly? Or perhaps trying to diagnose the wrong thing?

**Comment by sgpearson17** [comment]
Using this yaml file:
[config.example_soulsby.yaml](https://github.com/user-attachments/files/28843643/config.example_soulsby.yaml)

**Comment by jreyns** [comment]
> Using this yaml file: [config.example_soulsby.yaml](https://github.com/user-attachments/files/28843643/config.example_soulsby.yaml)

will try to debug; this is with inlet_sedtrails.nc? nv, yes it is

**Comment by jreyns** [comment]
> Hi Johan,
> 
> I tried running it with an assortment of polygons, intending to intercept the particle paths, but as far as I can tell there is no discernible difference in behaviour when I use them:
> 
> ```
> domain:
>   pol_file: ./outer_domain.pol
>   inner_boundary_pol_files:
>     - ./islands_small.pol
>   boundary_class_pol_files:
>     open:
>       - ./tiny_offshore_boundary_edges.pol
>     land:
>       - ./examples/land_bnd_inlet.pol
> ```
> 
> image
> 
> `outer_domain.pol`
> 
> ```
> L000001
>            5           2 
> 36200 10000
> 36200 19500
> 43800 19500
> 43800 10000
> 36200 10000
> ```
> 
> `islands_small.pol`
> 
> ```
> L000001
>            5           2
> 39800 17000
> 39800 17200
> 40400 17200
> 40400 17000
> 39800 17000
> ```
> 
> tiny_offshore_boundary_edges.pol
> 
> ```
> L000001
>            4           2 
> 39400 16200
> 39400 17600
> 40600 17600
> 40600 16200
> ```
> 
> `land_bnd_inlet.pol`
> 
> ```
> L000001
>            5           2
> 39000 15000
> 41000 15000
> ```
> 
> Am I running something incorrectly? Or perhaps trying to diagnose the wrong thing?

image


image

File: examples/inlet_sedtrails_example.yaml [hunk 1]
```diff
@@ -0,0 +1,89 @@
...
+#
+# Intended use from the examples directory:
+#   cd examples
+#   sedtrails run -c inlet_sedtrails_example.yaml
+#
+# The input file path below follows that working-directory convention.
+
+general:
+  preprocess: true
+  compute_pathways: true
```
sgpearson17, line 11: What is this compute_pathways line doing? it keeps getting hung up here when I try to run it? compute_pathways is not in the active schema- is this something new that you added or something old that found its way into this example?

examples/inlet_sedtrails_example.yaml
```diff
@@ -0,0 +1,89 @@
...
+  data: ../sample-data/inlet_sedtrails.nc
+  read_interval: 12H
+  repeat_eulerian_fields: true
+
+# The inlet sample is useful for testing boundary behavior. Leave these
+# commented until matching Tekal polygons are available next to this file.
+domain:
+  boundary_class_pol_files:
+    open:
+      - ./offshore.pol
```
sgpearson17, line 29: Something is not right- I cannot run this example without commenting out L26-29.  When I tried turning on `report_domain_exit_updates: true`, at least it correctly identifies nothing leaving, so that feature works:
```
2026-06-23 12:31:16,972 INFO sedtrails.simulation_orchestrator.simulation_manager: Particles left domain during run: 0/15
2026-06-23 12:31:16,973 INFO sedtrails.simulation_orchestrator.simulation_manager: Particles beached on land during run: 0/15
```

When I looked at it with copilot it says:
```
The failure is exactly the schema rule: boundary_class_pol_files is allowed only inside a valid domain definition that also sets [pol_file]
```

Which seems a bit strange, as this is the feature we are trying to add. is this an old/irrelevant example?

sgpearson17: I think something isn't entirely clear to me with regards to how these polygons should be defined, either- maybe we need a bit of a tutorial and/or documentation. I loaded all the pol files here and it looks to me like your `offshore.pol` file (the horseshoe-like one at the top) is a place where the particles should at least intersect:

Image

Is the idea that particles are removed when they enter an offshore polygon? that seems a bit counter intuitive to me...

In principle, we should see particles "leaving the domain" with this run and `offshore.pol' turned on- they're in the right place:

Image

examples/check_bnd/config.example_soulsby.yaml
```diff
@@ -0,0 +1,60 @@
...
+inputs:
+  data: ../../sample-data/inlet_sedtrails.nc
+  read_interval: 10D  # Time chunk size for reading input data
+  repeat_eulerian_fields: true  # Set true to loop forcing from the first input timestamp
+# Optional domain controls for islands/cutouts and boundary actions.
+# Tekal .pol paths are resolved relative to this configuration file.
+domain:
+  pol_file: ./outer_domain.pol
+  inner_boundary_pol_files:
+    - ./islands_small.pol
```
sgpearson17, line 15: this one seems to work better though:
```
2026-06-23 12:38:22,547 INFO sedtrails.simulation_orchestrator.simulation_manager: Particles beached on land during run: 6/10 (FineSand=6/10)
```
Particles seem to get stuck at the edges (approx in green) and then move away again, while the two particles seeded inside `islands_small.pol` stay where they are.
Image

For comparison, when I comment out those lines, particles travel right through that location since there is no land boundary:

Image

with 
```
2026-06-23 12:46:37,341 INFO sedtrails.simulation_orchestrator.simulation_manager: Particles left domain during run: 0/10
2026-06-23 12:46:37,341 INFO sedtrails.simulation_orchestrator.simulation_manager: Particles beached on land during run: 0/10
```

examples/check_bnd/config.example_soulsby.yaml
```diff
@@ -0,0 +1,60 @@
...
+    reference_date: 1970-01-01  # Default reference date for the input model
+    morfac: 1  # Morphological acceleration factor for time decompression
+inputs:
+  data: ../../sample-data/inlet_sedtrails.nc
+  read_interval: 10D  # Time chunk size for reading input data
+  repeat_eulerian_fields: true  # Set true to loop forcing from the first input timestamp
+# Optional domain controls for islands/cutouts and boundary actions.
+# Tekal .pol paths are resolved relative to this configuration file.
+domain:
+  pol_file: ./outer_domain.pol
```
sgpearson17, line 13: It's not entirely clear to me what role this plays- if I only specify this (and not the other open/land bnds below), I don't get any discernible effect, even when I use `examples\check_bnd\tiny_offshore_boundary_edges.pol`. What role does this play and how can I test that? Some documentation would be useful here.

examples/check_bnd/config.example_soulsby.yaml
```diff
@@ -0,0 +1,60 @@
...
+  inner_boundary_pol_files:
+    - ./islands_small.pol
+  boundary_class_pol_files:
+    land:
+      - ./land_bnd_inlet.pol
```
sgpearson17, line 18: ah interesting- for some reason it won't let me include an `open` boundary unless I also specify a `land` boundary- I suspect this may be related to the error I had with `inlet_sedtrails_example.yaml`

I now get 6 particles leaving the domain:
```
2026-06-23 13:03:02,892 INFO sedtrails.simulation_orchestrator.simulation_manager: Particles left domain during run: 6/10 (FineSand=6/10)
2026-06-23 13:03:02,893 INFO sedtrails.simulation_orchestrator.simulation_manager: Particles beached on land during run: 0/10
```
which seems correct:
Image

examples/inlet_sedtrails_example.yaml
sgpearson17: after revisiting this (see my comments for `config.example_soulsby.yaml`), I think the issue with this example is that the domain polygon and land polygon were not also specified. It didn't run with just open.

sgpearson17 (changes requested):
Hi Johan,

1. I went through this and it seems to work, but it took quite a bit of playing around before I could figure out what it was doing and how to test each of the different boundary types. I think some better documentation would help with this.  
2. Also, if you take a look at my comments, it seems that you need to specify the full domain polygon plus land + open, not just e.g. open. This should either be documented or the flexibility to run with only one of them added.
I will be doing more sedtrails tomorrow, so let me know if you want to have a chat, maybe in the morning?

sgpearson17:
Hang on I'm an idiot, I missed [simulation-params.md](https://github.com/sedtrails/sedtrails/pull/411/changes#diff-d88e52bce895ed07256885fd4d0df28df6c95881ecdaad2b81832fea3a6d9498)

sgpearson17:
Ok, I see the documentation there and it would have helped a bit, but I think that a sketch or a screenshot would be beneficial. Also, one thing that isn't clear to me is whether our offshore boundaries are meant to encompass the domain, or be a sort of "trap" polygon around the edge of the domain that particles can "fall into". What if users just specify a line segment? does it need to be a closed polygon?  And how is the domain polygon used in the absence of open/land bnds?

Maybe let's call tomrrow if you can?

docs/user/simulations.md
```diff
@@ -14,7 +14,11 @@ For a detailed reference of all available parameters, please refer to the [Simul
...
 
-For FM and SFINCS models, the optional `domain.inner_boundary_pol_files` setting can use Tekal polygons to mask islands/cutouts. For FM, SFINCS, and XBeach models, `domain.boundary_class_pol_files` can classify boundary crossings as `open` or `land`. See the domain section of the parameter reference for the full behavior.
+For FM and SFINCS models, the optional `domain.inner_boundary_pol_files` setting can use Tekal polygons to mask islands/cutouts. For FM, SFINCS, and XBeach models, `domain.boundary_class_pol_files` can classify boundary crossings as `open` or `land`. These boundary-class polygons work with either `domain.pol_file` or a `domain.subset_x`/`domain.subset_y` extent. They select active boundary edges by edge midpoint. Drawing rule: boundary-class polygons may be wider than a thin line as long as they only contain the intended edge midpoints, and do not include neighboring or unrelated boundary-edge midpoints. See the domain section of the parameter reference for the full behavior.
+
+![Domain polygon definition schematic](../_static/img/domain-polygon-definitions.png)
```
sgpearson17, line 21: I think this file should be uploaded too?

sgpearson17 (changes requested):
Docs are a bit clearer now. I think I will come back to this later when I review the docs before release. Missing the png? If you upload that too we can approve this.

docs/user/simulations.md
jreyns: This should be there? I have explicitly added it, and it shows up in my local sphinx build :/

sgpearson17:
Ah yes, it's there now. I think I had reviewed your first commit @jreyns
[Try to explain boundary polygons](https://github.com/sedtrails/sedtrails/pull/411/commits/831d1604023a6e50c822720130bb850a26d44cca)
[831d160](https://github.com/sedtrails/sedtrails/pull/411/commits/831d1604023a6e50c822720130bb850a26d44cca) which didn't yet have the figure included. I can see it now.

My only (final final) question is from looking at the figure- is the "regular closed boundary" defined by the edge of the "active domain from subset/pol_file"?

jreyns:
> is the "regular closed boundary" defined by the edge of the "active domain from subset/pol_file"

Yes, geometrically: the regular boundary is the edge of the active domain, and comes from:
- the input model grid, if domain is omitted
- the clipped/masked domain from domain.pol_file, if supplied
- the rectangular clipped domain from subset_x + subset_y, if supplied
plus any holes/cutouts created by inner_boundary_pol_files

After that is settled, the boundary_class_pol_files.open and .land only override parts of that active boundary by testing boundary-edge midpoints, and set the particle behaviour according to that.

sgpearson17:
Marvelous. I just added that bit of text to the docs below what you had already written, and will approve

sgpearson17 (approved):
Approved after many discussions

PR #428, opened 2026-06-19
Fix XBeach plugin for curvilinear grids with cutouts

### LINKED ISSUE #426: New XBeach version with curvilinear cutouts is not read correctly by the XBeach plugin
gridx and gridy have NaN values and crash the algorithm.

jreyns:
Input reader breaks on undefined grid coordinates in the XBeach nc flow field file

Closes #426


jreyns:
Needed for NCK summerschool. If it breaks, we will fix.

PR #429, opened 2026-06-21
Add a pan and zoom buttons to the seeder window

jreyns:
Allows to have more flexibility when selecting seeding locations in large models


jreyns:
Another NCK summer school small convenience commit. Will fix if it breaks something.

Note: the changes in the simulation manager to avoid the dask dependency need to be checked in the 'fast' branches

Issue #437, opened 2026-07-03
Connect Diffusion Library with Main Model Loop


sgpearson17:
When testing the passive tracer I realized that the diffusion library was not actually yet connected with the main model loop (for passive, vanwesten, or soulsby). Not sure how this one slipped through but I think we need to do it before we make a release. I can try next week.

Proposal from codex:

## Summary

Current behavior is deterministic advection for all tracer methods in the movement step.
The configuration key `particles.populations[].characteristics.diffusion_coefficient`
exists for passive particles, but diffusion is not wired into runtime advection.

This note proposes a unified diffusion feature for `passive_tracer`, `soulsby`, and
`vanwesten`, with a configurable strategy from
`src/sedtrails/particle_tracer/diffusion_library.py` and a Brownian default.

## Current state

- Diffusion implementations exist in `diffusion_library.py`.
- The simulation loop computes physics fields and then calls
  `ParticlePopulation.update_position(...)`.
- `update_position` advances with velocity fields only; no diffusion increment is applied.
- Soulsby and van Westen plugins produce transport velocities/probabilities and mixing
  fields, but do not apply random walk or any kh-based displacement.

## Goals

- Add optional, method-agnostic horizontal diffusion to particle motion.
- Keep backward compatibility by defaulting to no diffusion unless explicitly enabled,
  or use `brownian` with `coefficient: 0.0` as equivalent no-op.
- Allow per-population diffusion setup (strategy + coefficient) across all tracer methods.
- Keep deterministic reproducibility via explicit RNG seed handling.

## Non-goals

- No anisotropic tensor diffusion in this phase.
- No vertical diffusion or turbulence closure model coupling in this phase.
- No changes to existing transport-probability logic.

## Proposed configuration

Add a new optional top-level key under each population:

```yaml
particles:
  populations:
    - name: population_1
      particle_type: passive
      diffusion:
        method: brownian
        coefficient: 100.0
        seed: 42
```

### Configuration semantics

- `diffusion.method`:
  - allowed: `none`, `brownian`, `random`, `gradient`
  - default: `brownian`
- `diffusion.coefficient`:
  - scalar horizontal diffusivity kh in m2/s
  - default: `0.0`
  - must be `>= 0`
- `diffusion.seed`:
  - optional integer
  - when present, diffusion draws are reproducible for that population

### Backward compatibility

- Keep accepting legacy `characteristics.diffusion_coefficient`.
- Migration rule:
[...]

PR #439, opened 2026-07-13
Add MacDonald Q3D particle tracking and timestep-aware physics to the latest dev (refactored) SedTRAILS version

the-vdag:
This PR adds the MacDonald particle tracking implementation, including Q3D particle motion, timestep-dependent physics, entrainment/deposition handling, and diagnostic outputs.

Main changes:

- Added a new `macdonald` physics plugin based on MacDonald (2006).
- Added MacDonald 2D/Q3D support to the simulation runtime plan and physics converter.
- Split MacDonald Q3D physics into:
  - base physics computed before CFL timestep selection,
  - timestep-dependent physics computed after the active timestep is known.
- Updated CFL timestep handling so the CFL timestep is bounded by the configured simulation timestep (.yaml input) and input-data timestep (sedtrails_data).
- Added Q3D particle position updates with:
  - entrainment from the bed,
  - suspended/deposited/buried status tracking,
  - horizontal random-walk diffusion,
  - vertical advection, settling, and diffusion,
  - boundary and beaching handling.
- Added selectable entrainment probability laws:
  - `poisson`
  - `linear`
- Added `selected_shear_velocity` support for transport-related velocity calculations.
- Added first-substep Q3D diagnostics and NetCDF output fields with explicit `first_substep_*` names.
- Extended YAML schemas for MacDonald/Q3D configuration options.
- Added roughness/rouse diagnostic resources (supporting files within folder resources)


sgpearson17:
Awesome, thanks for all your efforts on this, Vassia! I know this is the culmination of a ton of work. It seems some of the pytests are still failing. Let's take a look at this PR together on Wednesday morning if you're still here. You can walk me through some of the changes and we can investigate the pytests if there are still problems. I will try to finish my review on this wednesday so that we can merge into dev before my holidays.

sgpearson17:
Thanks for all your hard work on this, Vassia! A few points as discussed:
1. Some check plots of zp vs x and/or t would be helpful to make sure the particles are doing what we think they are doing
2. If you can provide some documentation (auto-generated is fine for now) explaining the bits that you've added, that would be helpful. In this documentation I think we should also clearly state that this Q3D stuff is in beta mode (even after merging with the dev branch) as part of your ongoing PhD research to give you a bit of breathing room and ensure that other people don't start using it just yet, until after you have had a chance to publish on it.
3. If you can also share your powerpoint slides with Johan and I, then we can refer to that as needed when reviewing the code

examples/sedtrails-example-macdonald.yaml
sgpearson17, line 1: can you add to the user docs to say something explicitly about the keywords needed to run macdonald and generally what they do? this can be auto-generated as a start

resources/create_lookuptable_zs_over_h_rouse.py
sgpearson17, line 1: it would be good to add a unit test for this stuff, to check that values retrieved from the lookup table match an independently-calculated z value using the desired input values. Even better if you can tie it back to a specific number given as an example in the macdonald report, though I can't recall if they mention one or not.

resources/diagnostic_velocity_profile_roughness_switch_d50_300um_uc_0.5_uw_0.0_phi_0.png
sgpearson17, line 1: This and the other images should probably not be included here, but certainly keep them as input for your paper. @jreyns - how should we approach this sort of thing more generally? i.e. sanity-check images produced during the development process. Should some of this material become unit tests? or at least be reproducible somewhere?

resources/explore_roughness_coefficients.py
sgpearson17, line 1: I also feel like we shouldn't include this here but I also don't think we should get rid of it entirely.  
@jreyns- what should our philosophy be on this sort of thing. I also have e.g. python notebooks with my own tests and verification stuff and I don't know how much we should keep/share and store here vs elsewhere. what is best practice?

resources/macdonald_zs_over_H_vs_rouse.png
sgpearson17, line 1: This and the other images should also be removed, but the code that you used to generate them should still be available; let's see what Johan says about the other stuff in `resources`

src/sedtrails/config/population.schema.json
```diff
@@ -89,6 +89,87 @@
...
+        "macdonald_method": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "flow_field_name": {
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    },
+                    "description": "flow field name for van Westen method used in particle computation"
```
sgpearson17: van Westen --> MacDonald?

src/sedtrails/config/population.schema.json
```diff
@@ -89,6 +89,87 @@
...
+                },
+				"max_suspended_velocity_factor": {
+				  "type": ["number", "null"],
+				  "default": 1.0,
+				  "description": "Optional cap on suspended-load advection velocity, expressed as a multiple of the depth-averaged flow velocity. If null, no capping is applied."
+				},
+				"q3d_horizontal_diffusion_enabled": {
+				  "type": "boolean",
+				  "default": true,
+				  "description": "If false, disable Q3D horizontal turbulent diffusion and its random-walk velocity components without changing geometric-scheme vertical diffusion."
```
sgpearson17, line 198: @jreyns - let's discuss how this fits with the general issue of the diffusion library not currently being coupled to the rest of the model (#437)

tests/particle_tracer/test_particle_seeder.py
sgpearson17, line 1: when I tried running the default example with vanwesten using your PR code, I get an error:
`Error running simulation: 'ParticlePopulation' object has no attribute 'update_bed_level_change_after_movement' `
The breakage is in `simulation_manager.py:1528`. In the vanwesten mode, it calls `population.update_bed_level_change_after_movement(bed_level)`, but the current `ParticlePopulation` implementation in `particle_seeder.py:960` does not define that method anywhere. 

I think that method is designed to avoid the "tunnelling" behaviour that we previously observed.

Apparently the skipped tests in `test_particle_seeder.py` show that this behavior was expected but not actively exercised, so the regression slipped through.

These tests should not be skipped, and your code should instead be written such that it can pass the tests. I would also like to see more tests for the new features that you added, such that we can make sure the behaviour is correct and to make sure that future PRs don't break it.

sgpearson17: On a similar note, when running `config.example_soulsby.py` I get the error: 
`Error running simulation: type object 'PhysicsPlugin' has no attribute 'safe_divide'`
I think this was also in your failed unit test?

src/sedtrails/data_manager/netcdf_writer.py
sgpearson17, line 1: check that your q3d fields are only saved if the correct option is enabled

tests/particle_tracer/test_particle_seeder.py
sgpearson17: See `tests\transport_converter\plugins\test_vanwesten_plugin.py`, `test_physics_plugin.py`, or `test_soulsby_plugin.py` for examples of the kind of things to include in the tests. Co-pilot is very helpful for this sort of thing!

src/sedtrails/simulation_orchestrator/simulation_manager.py
sgpearson17: this is the line where I got errors when running the default vanwesten example `examples\sedtrails-example.yaml` using your PR branch. it would be good to check if there is any conflict or redundancy between this method in his code (see also `update_burial_depth` in 'particle_seeder.py`)

src/sedtrails/particle_tracer/particle_seeder.py
```diff
@@ -1190,56 +1272,1556 @@ def _interpolate_particle_fields(self, fields):
...
+            0.0,
+            np.where(
+                z_p <= z_c_safe,
+                deficit_coefficient * u_zp,
+                np.where(z_p <= 1.4 * z_c_safe, u_transition, u_zp),
+            ),
+        )
+
+    @staticmethod
+    def _turbulent_diffusion_coefficients(
```
sgpearson17: Should we move some of this out to the physics library when we resolve #437?

src/sedtrails/particle_tracer/particle_seeder.py
```diff
@@ -1190,56 +1272,1556 @@ def _interpolate_particle_fields(self, fields):
...
+            candidate_z = z_p_old - np.maximum(np.nan_to_num(settling_velocity, nan=0.0), 0.0) * dt
+            z_p_new = np.maximum(candidate_z, z_c_new)
+        elif scheme == 'rouse_profile':
+            z_p_new = self._sample_rouse_profile_height(water_depth_new, rouse_number_new, rng=rng)
+        else:
+            raise ValueError(f'Unsupported q3d vertical update scheme {scheme!r}')
+
+        return np.clip(np.nan_to_num(z_p_new, nan=0.0, posinf=0.0, neginf=0.0), 0.0, water_depth_new)
+
+    def update_q3d_particle_position(
```
sgpearson17: There is a lot of new code in particle_seeder in this PR, but I'm not sure all of it needs to go here- could you maybe refactor some of this into a separate function? (especially this part, but also including some of the functions above)

src/sedtrails/transport_converter/plugins/physics/macdonald.py
```diff
@@ -0,0 +1,1322 @@
...
+            particle_velocity_over_flow = PhysicsPlugin.safe_divide(centroid_particle_velocity, flow_velocity_magnitude)
+            log_law_argument = PhysicsPlugin.safe_divide(30 * z_s, profile_roughness_height)
+            with np.errstate(divide='ignore', invalid='ignore'):
+                suspended_load_velocity_lnpart = np.where(
+                    log_law_argument > 0,
+                    np.log(log_law_argument),
+                    np.nan,
+                )
+
+            sedtrails_data.add_physics_field('max_shields_number', max_shields_number)
```
sgpearson17, line 285: Are all of these fields always necessary, or can they be selectively added as needed, depending on the options chosen? I just wonder if this will suck up a lot of memory for larger simulations...

src/sedtrails/transport_converter/plugins/physics/macdonald_sandbox.py
sgpearson17, line 1: delete this?

sgpearson17 (changes requested):
Great work on all this, Vassia! there are a bunch of things to address further but I think this makes a really valuable contribution to sedtrails as a whole. I will discuss some of this with Johan tomorrow.  

Note- if you address some of the comments, whether mine or copilot's, can you click on the 'resolved' button once you're done? then we can better keep track of everything.

examples/sedtrails-example-macdonald.yaml
the-vdag: [SedTRAILS_MacDonald_2D_Q3D_Technical_Manual.docx](https://github.com/user-attachments/files/30414045/SedTRAILS_MacDonald_2D_Q3D_Technical_Manual.docx)

This is an autogenerated technical manual and includes an overview of the yaml keywords

src/sedtrails/transport_converter/plugins/physics/macdonald.py
```diff
@@ -0,0 +1,1318 @@
...
+            module_dir = os.path.dirname(os.path.abspath(__file__))
+
+            # file name for the NEW 1D table
+            lookup_path = os.path.join(module_dir, "macdonald_z_over_h_lookup.nc")
+
+            if not os.path.isfile(lookup_path):
+                raise FileNotFoundError(
+                    f"MacDonald lookup table not found: {lookup_path}\n"
+                    "Expected 'macdonald_z_over_h_lookup.nc' to be in the same folder as this plugin file."
+                )
```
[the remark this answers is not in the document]
the-vdag: The file is now git tracked and present in the repo.

src/sedtrails/transport_converter/plugins/physics/soulsby.py
```diff
@@ -272,3 +272,11 @@ def add_physics(self, sedtrails_data: SedtrailsData, grain_properties: dict[str,
...
         sedtrails_data.add_physics_field('soulsby_b', soulsby_b)
         sedtrails_data.add_physics_field('mixing_layer_thickness', mixing_layer_thickness)
+        sedtrails_data.add_physics_field('suspended_velocity', Rs*flow_velocity_magnitude) # this is for debugging purposes
+        sedtrails_data.add_physics_field('bedload_velocity', bed_load_velocity) # this is for debugging purposes
+        sedtrails_data.add_physics_field('rouse_number', rouse_number) # this is for debugging purposes
+        sedtrails_data.add_physics_field('shear_velocity_ratio', max_shear_velocity/mean_shear_velocity) # this is for debugging purposes
+        sedtrails_data.add_physics_field('suspended_velocity_over_da_velocity', Rs) # this is for debugging purposes
+        sedtrails_data.add_physics_field('max_shear_velocity', max_shear_velocity) # this is for debugging purposes
+        sedtrails_data.add_physics_field('mean_shear_velocity', mean_shear_velocity) # this is for debugging purposes
+        sedtrails_data.add_physics_field('particle_velocity_over_da_velocity', PhysicsPlugin.safe_divide(grain_velocity_magnitude, flow_velocity_magnitude, fill=0.0)) # this is for debugging purposes
```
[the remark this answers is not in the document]
the-vdag: Extra diagnostics are gated and safe divide has been implemented

src/sedtrails/transport_converter/physics_lib.py
```diff
@@ -472,16 +524,34 @@ def compute_grain_properties(
     Soulsby, R. L., & Whitehouse, R. (1997). Threshold of sediment motion in coastal environments.
     Pacific Coasts and Ports '97: Proceedings of the 13th Australasian Coastal and Ocean
     Engineering Conference and the 6th Australasian Port and Harbour Conference; Volume 1. Equation 14
+
+    MacDonald, N., Davies, M., Zundel, A., Howlett, J., Demirbilek, Z.,
+    Gailani, J., Lackey, T., & Smith, J. (2006). *PTM: Particle Tracking Model. 
+    Report 1: Model Theory, Implementation, and Example Applications*.      '
+    U.S. Army Corps of Engineers. Equation 28
```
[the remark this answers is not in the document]
the-vdag: Cleaned!

src/sedtrails/transport_converter/plugins/physics/macdonald.py
```diff
@@ -0,0 +1,1318 @@
...
+import numpy as np
+import xarray as xr
+from scipy.spatial import cKDTree
+# from random import random
+from sedtrails.transport_converter import physics_lib
+from sedtrails.transport_converter.plugins import BasePhysicsPlugin
+from sedtrails.transport_converter import SedtrailsData
+
+
+class PhysicsPlugin(BasePhysicsPlugin):  # all clases should be called the PhysicsPlugin
```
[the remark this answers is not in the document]
the-vdag: fixed

src/sedtrails/transport_converter/plugins/physics/macdonald.py
```diff
@@ -0,0 +1,1318 @@
...
+                n_invalid = int(np.count_nonzero(invalid))
+                n_low = int(np.count_nonzero(low_oob))
+                n_high = int(np.count_nonzero(high))
+                print(
+                    "[MacDonald lookup] Some inputs were invalid or out of lookup range.\n"
+                    f"  invalid (NaN/Inf/rouse<=0/depth<0): {n_invalid}/{n_tot}\n"
+                    f"  rouse < rmin ({rmin:g}) -> NaN: {n_low}/{n_tot}\n"
+                    f"  rouse >= {R_ASYM:g} -> asymptote z_s/H={ASYM:.6g}: {n_high}/{n_tot}\n"
+                    f"  lookup rouse range: [{rmin:g}, {rmax:g}] from {PhysicsPlugin._macdonald_lookup_path}"
+                )
```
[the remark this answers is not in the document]
the-vdag: fixed

resources/create_lookuptable_zs_over_h_rouse.py
the-vdag: Went over the calculation, found a bug in the formula used to caclulate the lookup table (wrong parenthesis). Created (with Claude) the unit tests (1 that checks calculations using the lookup table against the full formula, one that checks the value at R=1 when ln1=0, and a few more to check the asymptote behaviour for R>20 ( for very big R, tanh(R->oo) goes to 1).

src/sedtrails/config/population.schema.json
the-vdag: fixed

src/sedtrails/transport_converter/plugins/physics/macdonald_sandbox.py
the-vdag: deleted

the-vdag: deleted

src/sedtrails/pathway_visualizer/sedtrails_plotting.py
```diff
@@ -286,44 +286,44 @@ def plot_trajectories_by_baseline(
...
+    Returns
+    -------
+    matplotlib.axes.Axes
+        Axes containing the trajectory scatter plot.
+
+    Notes
+    -----
+    Mirrors the rotated-axis idea used in the MATLAB pathway plots. The color
+    is computed from the *initial* rotated-X coordinate per particle and then
+    broadcast to all timesteps. :contentReference[oaicite:2]{index=2}
```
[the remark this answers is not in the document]
the-vdag: removed

src/sedtrails/pathway_visualizer/sedtrails_plotting.py
```diff
@@ -689,21 +689,21 @@ class ParticleStats:
...
-        First timestep considered stable for plotting or analysis.
-
-    Returns
-    -------
-    List[ParticleStats]
-        List containing the computed records.
-    """
+    """
+    Compute per-particle trajectory statistics similar to analyze_pathways.m. :contentReference[oaicite:3]{index=3}
+
```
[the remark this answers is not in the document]
the-vdag: removed

src/sedtrails/transport_converter/plugins/physics/macdonald.py
```diff
@@ -0,0 +1,1318 @@
...
+
+    # Shared class-level cache for MacDonald lookup table
+    # Loaded lazily on first use and shared across all plugin instances
+    _macdonald_lookup_da = None
+    _macdonald_lookup_path = None
+    _macdonald_warned_oob = False
+
+    def __init__(self, config, tracer_methods: None):
+        super().__init__()
+        self.config = config
```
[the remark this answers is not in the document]
the-vdag: fixed

src/sedtrails/particle_tracer/particle_seeder.py
```diff
@@ -381,11 +407,45 @@ def __post_init__(self):
         _burial_depth = find_value(self.population_config, 'seeding.burial_depth', {})
         if not _burial_depth:
             raise MissingConfigurationParameter('"burial_depth" is not defined in the population configuration.')
+        if not isinstance(_burial_depth, dict):
+            raise TypeError('"seeding.burial_depth" must be a dictionary.')
         self.burial_depth = _burial_depth
```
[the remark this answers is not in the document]
the-vdag: Addressed

src/sedtrails/transport_converter/plugins/physics/macdonald.py
the-vdag: Added export_diagnostic_fields (default false) to gate the ~21 unused fields; the ~9 actually-consumed fields stay always-on. Where a "diagnostic" field's value was secretly needed elsewhere (e.g. bed_load_velocity), we keep computing it, we only skip the extra storage, which is what actually costs memory (add_physics_field stores a reference, not a copy). A test locks in that required field values are bit-identical regardless of the flag.

Also consolidated: merged q3d_export_grid_diagnostics into the new flag, and dropped the legacy q3d_diagnostics string (always redundant with q3d_save_first_substep_diagnostics in real configs).

src/sedtrails/particle_tracer/particle_seeder.py
the-vdag: Made a new file (src/sedtrails/particle_tracer/q3d_macdonald_motion.py
 ,Mixing class) to hold the Macdonald specific methods. Did not change anything in the methods themselves.

resources/explore_roughness_coefficients.py
jreyns: @the-vdag @sgpearson17 if we keep it, it should be in the docs folder I guess. We have to avoid that that one get cluttered though, and keep all the RTD stuff in a separate subfolder. This then can go into 'technical documentation' or similar?

the-vdag:
Check internal use of multiple fractions (e.g. qs_qt) @jreyns

src/sedtrails/particle_tracer/q3d_macdonald_motion.py
```diff
@@ -242,7 +242,9 @@ def initialize_macdonald_2d_release_state(self) -> None:
 
         for mode in np.unique(modes[to_initialize]):
             idx = to_initialize & (modes == mode)
-            if mode in {'height_above_bed', 'centroid_on_release'}:
```
the-vdag: For initializing particles in the 2D model we only require that both
'height_above_bed' and 'centroid_on_release' yield a positive
height_above_bed so the particle is treated as suspended. Distinguishing
the two modes here is therefore not strictly necessary for suspension/deposition
determination — they could share the same branch as long as the result is > 0.0.

That said, the explicit per-mode handling below is more complete and clearer:
   - it documents intent (e.g. centroid_on_release -> height = 1.0),
   - it makes NaN/negative handling explicit for height_above_bed,
   - and it keeps room for mode-specific logic later if needed.

src/sedtrails/simulation_orchestrator/simulation_manager.py
```diff
@@ -847,9 +847,15 @@ def _should_update_bed_level_after_movement(tracer_plan) -> bool:
         - Also run for any tracer configured with ``no_probability`` so
           particle ``z`` follows bed level after movement (passive/soulsby included).
 
-        Future quasi-3D tracers can introduce exceptions here when vertical
-        position is solved independently from bed level.
+        MacDonald Q3D is excluded because it solves and stores its own absolute
```
the-vdag, line 850: not sure about this. It seems that Macdonald q3D is exlcuded from checking and applying bed level change, this is ok for now, but should be revisited when handling morphodynamic input for Macdonald.

src/sedtrails/particle_tracer/q3d_macdonald_motion.py
```diff
@@ -184,13 +185,15 @@ def _turbulent_diffusion_coefficients(
...
+        # Depth-scaled variant of equations 49 and 50. In the report K_Ev
+        # carries length units; here K_Ev is dimensionless and h supplies that
+        # length explicitly.
         vertical = np.zeros_like(water_depth, dtype=float)
         if compute_vertical:
             shape = np.zeros_like(water_depth, dtype=float)
             valid = water_depth > 0.0
             shape[valid] = z_p[valid] * (water_depth[valid] - z_p[valid]) ** 2 / water_depth[valid] ** 3
-            vertical = M_b * K_Ev * flow_velocity_magnitude * shape
+            vertical = M_b * K_Ev * water_depth * flow_velocity_magnitude * shape
```
the-vdag, line 196: Aha, thats interesting - Equation 49 in Macdonald misses the waterdepth and thus gives a vertical Ev with m/s while it should be m2/s so that u_D (eq 51) ends up being m/s. Good work spotting this!

src/sedtrails/particle_tracer/q3d_macdonald_motion.py
the-vdag: ok - got it. Makes sure that if height_above_bed is 0.0 those particles are initialised on the bed

src/sedtrails/transport_converter/plugins/physics/macdonald.py
```diff
@@ -234,7 +234,11 @@ def add_physics(
         #Question Vassia: should the max_suspended_velocity_factor be used here or in the final step - on the mean_particle_velocity?
         
         # bed load velocity (MacDonald et al., 2006, equation 30) - Engelund & Fredsoe (1976), same as Soulsby et al (2011)
-        bed_load_velocity = physics_lib.compute_bed_load_velocity(max_shields_number, critical_shields, max_shear_velocity)
+        bed_load_velocity = physics_lib.compute_bed_load_velocity(
+            max_shields_number,
+            critical_shields,
+            mean_shear_velocity,
```
the-vdag, line 246: Why mean_shear_velocity and not selected_shear_velocity ?? For model_native sediment transport runs these two fields are different!

src/sedtrails/simulation_orchestrator/simulation_manager.py
```diff
@@ -942,7 +942,7 @@ def _macdonald_2d_deposition_parameters(
                 ),
                 'shear_velocity_field': retriever.get_scalar_field_bounds(
                     field_time_seconds,
-                    'max_shear_velocity',
```
the-vdag: Agreed.

the-vdag:
Went over Johan's fixes: 

- [x] fix: preserve MacDonald transition fields - ok!
- [x] fix: preserve MacDonald Q3D vertical position - ok for now, It seems that Macdonald q3D is exlcuded from chrecking and applying bed level change- to revisit when handling morphodynamic input for Macdonald.
- [x] fix: broadcast MacDonald settling velocity -ok!  
- [x] fix: differentiate Q3D surface over input time -ok! 
- [x] fix: use registered velocity in MacDonald example -ok!
- [x] fix: validate implemented MacDonald modes -ok! 
- [x] fix: restore Particle mobility compatibility -ok!
- [x] perf: cache MacDonald Q3D divergence - **proposed changes - cache only the geometry dependent stencils (neighbour indices and wights) keyed by the flow-field geometry (x,y,k, r_max). Then on every call apply the cached stencils to the current u,v to compute dudx,dvdy,divU. This way the velocity is always updated while we reuse the expensive geometry work.  This is what my last commit is about. Check if there is now a more efficient way for this using simplex/simplices**
- [x] fix: derive particle speed from local velocity -ok!
- [x] perf: reuse Q3D particle interpolation geometry -ok! uses previous simplices as a warm start to update the triangles the particles are in.
- [x] fix: separate mobility and advection shear - diffusion shear is the selected one (not the max by default) - also bed load velocity (advection) uses the mean shear velocity (maximum shields number is still used for the threshold check) -ok!
- [x] fix: align model-native profile with Chezy flow -ok!
- [x] fix: make Q3D vertical diffusivity dimensional -ok!  Aha, thats interesting - Equation 49 in Macdonald misses the waterdepth and thus gives a vertical Ev with m/s while it should be m2/s so that u_D (eq 51) ends up being m/s. Good work spotting this!
- [x] fix: reject layered velocity as depth averaged -ok!
- [x] fix: honor MacDonald 2D release height -ok! makes sure that if height_above_bed is 0.0 those particles are initialised on the bed
- [x] fix: broadcast equilibrium bedform inputs -ok!
- [x] fix: require MacDonald flow field -ok!
- [x] fix: validate vertical position values -ok! 
- [x] perf: cache MacDonald lookup arrays -ok!
- [x] chore: remove generated analysis artifacts -ok!
- [x] chore: normalize plotting source line endings -ok!
- [x] chore: clean branch whitespace -ok!
- [x] fix: reject static layered FM velocity -ok!
- [x] fix: expose Q3D horizontal diffusion factor -ok!
- [x] fix: reserve maximum shear for mobilisation -ok! yes but what do we use for bedload_velocity? mean_shear_velocity or selected_shear_velocity?? Commented on that on a previous commit
- [x] fix: normalize Chezy log profile depth average -ok!
- [x] fix: preserve Q3D surface-change gradient -ok!
- [x] fix: refresh Q3D shear between substeps -ok!
- [x] fix: handle Q3D exit before mobility finalization -ok!
- [x] fix: reject unsupported MacDonald burial modes -ok! 
- [x] fix: initialize Q3D state before first output -ok!
- [x] fix: preserve Q3D state across restart -ok!
- [x] fix: scope Q3D diagnostics per population -ok!
- [x] perf: cache only Q3D divergence - ok!
- [x] perf: avoid stacking interpolation fields - ok!
- [x] perf: share immutable runtime physics arrays -ok!
- [x] fix: initialize Q3D from available centroid -ok! 
- [x] fix: persist Q3D initialization state - ok!
- [x] fix: squeeze singleton FM velocity layer -ok!
- [x] fix: exclude unreleased particles from restart -ok!
- [x] fix: use current shear for Markov settling -ok!
- [x] fix: bound MacDonald centroid for Q3D - ok! so the centroid can not be above h/1.4=0.715h. Rouse profile gives max zc=0.5 h.

Issue #333, opened 2025-09-24
Fix update_burial_depth()


bartvanwesten:
Currently, first 'z' is computed using 'bed_level' (=updated via Eulerian input) and 'burial_depth'.
Afterwards, 'burial_depth' updated again.

This seems to work when particles don't move. However, if particles move over the bed, and the bed level differs over space, this method could fail. Only way is updating burial_depth based on bed_level_change, which should be computed explicitely (somewhere bed_level_prev should be stored).

sgpearson17:
**Note: this is the "tunneling" problem.** This can be found in ``particle_tracer/particle_seeder.py``

We need to separate between bed level changes in space and over time. If we only account for the bed level changes at t=t, then we run the risk that particles "tunnel" and get buried (i.e. burrowing into a hill instead of climbing up it)

The vertical position is computed by taking ``bed level - burial depth``, so if a particle moves "uphill" it may get buried instead.

Therefore in a previous timestep, you need to compute the bed level change instead of just the bed level.

Storing the previous bed level state requires a bit of thinking for where to do it, which is why Bart didn't finish at that time. Think of how to do this cleverly.

PR #440, opened 2026-07-14
Fix burial-depth spatial leak: re-sample bed level inside the flow-field loop

bartvanwesten:
## Summary

Particles in long runs progressively acquire spurious burial depth and become immobile — even on a **static bed**. In a year-long morphostatic Westerschelde run (1M particles, vanwesten tracer), 81% of particles had nonzero burial after one day and ~99% were flagged buried/immobile by year end, on a bed that never changes. This PR restores a previously-fixed invariant that was silently lost in a merge, and adds a regression test that guards it.

This also explains the earlier observation by @sgpearson17 that *"particles get buried and stop"* in a non-morphodynamic run.

## The invariant

`update_burial_depth()` books `bed_level - bed_level_previous` as a **temporal** bed change (zero on a static bed). That is only valid if `bed_level_previous` holds the bed at the particle's *current* position — which requires `update_bed_level_change_after_movement()` to run after **every** `update_position()`, i.e. **inside** the per-flow-field loop: the vanwesten tracer moves particles once per flow field (bed load + suspended), and the next flow field's `update_information()` re-interpolates the bed at the post-move positions.

## How it regressed

- `bdf7a99` (2026-06-02): invariant established — the re-sample call sits inside the flow-field loop.
- `90d0b5a` (2026-06-12, merge of upstream/dev into `particle_burial`): upstream had re-indented the main loop one level deeper; the merge resolution kept the re-sample block at its old indentation, which in the new structure sits **after** the loop. No line of the fix changed, so the regression was invisible in the merge diff, and no test guarded the invariant.

Since then, for two-flow-field tracers, every timestep booked `bed(post-bedload-move) - bed(pre-move)` — a **spatial gradient** term — as burial change. The `max(burial, 0)` floor rectifies this zero-mean noise into a monotonic ratchet (a reflected random walk), so burial grows ∝√t until it crosses the local mixing depth and the particle is flagged buried.

## The fix

Move the re-sample block back inside the flow-field loop (re-indent, plus a comment explaining why it must live there).

## Regression test

`tests/simulation_orchestrator/test_burial_static_bed.py`: an end-to-end vanwesten run on a synthetic static sloped bed with steady transport — particles must move while burial stays exactly zero. Verified both ways: **fails on pre-fix code** (4 mm spurious burial per simulated hour, matching the leak mechanism quantitatively) and passes with the fix.

## Validation

Re-ran three year-long Westerschelde production runs (1M particles) with the fix: the morphostatic run now keeps burial depth at exactly 0 for the full year, and the morphodynamic/observed-burial variants show slow, spatially coherent burial instead of the week-one ratchet.

### DIFF
--- a/src/sedtrails/simulation_orchestrator/simulation_manager.py
+++ b/src/sedtrails/simulation_orchestrator/simulation_manager.py
@@ -1406,9 +1406,19 @@ class Simulation:
                             timer.current_timestep,
                         )
 
-                    if tracer_plan.method_name == 'vanwesten':
-                        with self._profile_section('update_bed_level_after_movement'):
-                            population.update_bed_level_change_after_movement(bed_level)
+                        # Re-sample bed level at the new positions after EVERY
+                        # update_position. This must stay INSIDE the flow-field
+                        # loop: the next update_information copies bed_level to
+                        # bed_level_previous, and update_burial_depth treats the
+                        # difference as a temporal bed change. If the re-sample
+                        # runs only once per timestep, the second flow field's
+                        # update pairs a post-move bed with a pre-move previous,
+                        # leaking the spatial bed gradient into burial_depth
+                        # (see commit message: regression of bdf7a99 via merge
+                        # 90d0b5a).
+                        if tracer_plan.method_name == 'vanwesten':
+                            with self._profile_section('update_bed_level_after_movement'):
+                                population.update_bed_level_change_after_movement(bed_level)
 
                 # Update dashboard if enabled
                 if dashboard_update_due and dashboard_flow_field is not None:
--- a/src/sedtrails/simulation_orchestrator/simulation_manager.py
+++ b/src/sedtrails/simulation_orchestrator/simulation_manager.py
@@ -1420,6 +1420,10 @@ class Simulation:
                         # (see commit message: regression of bdf7a99 via merge
                         # 90d0b5a).
                         if tracer_plan.method_name == 'vanwesten':
+                            # note, maybe this should be moved out of this if 
+                            # block because other methods 
+                            # like soulsby or passive should also be able to 
+                            # navigate a morphodynamic input model
                             transport_probability_method = tracer_plan.transport_probability_method
                             if transport_probability_method != 'no_probability':
                                 with self._profile_section('update_bed_level_after_movement'):
--- a/src/sedtrails/simulation_orchestrator/simulation_manager.py
+++ b/src/sedtrails/simulation_orchestrator/simulation_manager.py
@@ -816,6 +816,23 @@ class Simulation:
         return sample_time + tolerance >= next_output_time or sample_time + tolerance >= end_time
 
     @staticmethod
+    def _should_update_bed_level_after_movement(tracer_plan) -> bool:
+        """Return whether post-move bed-level resampling should run.
+
+        Rules for current 2D workflows:
+        - Always run for ``vanwesten`` to preserve burial-depth bookkeeping.
+        - Also run for any tracer configured with ``no_probability`` so
+          particle ``z`` follows bed level after movement (passive/soulsby included).
+
+        Future quasi-3D tracers can introduce exceptions here when vertical
+        position is solved independently from bed level.
+        """
+        return (
+            tracer_plan.method_name == 'vanwesten'
+            or tracer_plan.transport_probability_method == 'no_probability'
+        )
+
+    @staticmethod
     def _initialize_population_output_status(populations, current_time: int | float) -> None:
         """Populate required status arrays before the initial trajectory sample is written."""
         for population in populations:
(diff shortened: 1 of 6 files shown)


sgpearson17:
Hi Bart,

Good catch- I noticed a problem when I was playing with the stochastic vs no-probability settings last week, but did not share my results yet.  To check, I first ran [sedtrails-example-no-prob-burial.yaml](https://github.com/user-attachments/files/30004851/sedtrails-example-no-prob-burial.yaml), which features two identical populations of sediment released at identical locations and tracked using `vanwesten`. Population 0 (red) uses `stochastic_transport` and Population 1 (grey) uses `no_probability`. 
image
If we look at the above figure, it seems like the grey (no probability) population is trapped against the edge of the island. The particles show fairly large (O(1km) excursions) which makes sense, except that the particles should not necessarily stay trapped like that. When I dug deeper it turned out that the particles were "buried" in a side wall but still moved around by the currents above because they are not `status_buried=1`. When you plot the mean burial depth of the particles in the above simulation, we see that the `no_probability` particles are all several meters deep but still mobile. 0% of the particles are officially buried, so they can still move around. On the other hand, the stochastic simulations are genuinely buried and many stop moving.
image
Is this a remnant of the old "tunneling" problem? Is that partly what you are addressing here?

Let me know if you want to chat further!

Stuart

sgpearson17:
NB I was doing the above checks here: https://github.com/sgpearson17/sedtrails-sgp/tree/check-no-prob-burial it's not very organized as I hadn't planned on sharing yet but figured it was relevant to mention now.

sgpearson17:
Also worth mentioning: the problem above went away when I added island boundary polygons to the test file.

sgpearson17:
Update: I ran the earlier-attached yaml (also below) with your PR branch here and still see the same issue, so I think that your fix here and the issue I noticed with `no_probability` are separate problems?

particle_trajectories

```
general:
  input_model: 
    format: fm_netcdf
    reference_date: 1970-01-01  # Default reference date for the input model
    morfac: 1  # Morphological acceleration factor for time decompression
inputs:
  data: ./sample-data/inlet_sedtrails.nc
  read_interval: 10D  # Time chunk size for reading input data
  repeat_eulerian_fields: true  # Repeat forcing after a valid start inside the input time range
# Optional domain controls for islands/cutouts and boundary actions.
# Tekal .pol paths are resolved relative to this configuration file.
# domain:
#   pol_file: ./outer_domain.pol
#   inner_boundary_pol_files:
#     - ./islands.pol
#   boundary_class_pol_files:
#     open:
#       - ./offshore_boundary_edges.pol
#     land:
#       - ./coastline_boundary_edges.pol
time:
  start:  2016-09-21 19:20:00
  timestep: 60S
  duration: 2D
  cfl_condition: 0.7  # CFL condition for adaptive timestep (0 = disabled)
particles:
  populations:
    - name: pop1_stochastic
      particle_type: sand
      characteristics:
        grain_size: 0.00025 
        density: 2650.0  
      tracer_methods:
        vanwesten:
          flow_field_name: 
            - bed_load_velocity
            - suspended_velocity
      transport_probability: stochastic_transport  # Options: no_probability, stochastic_transport, reduced_velocity
      seeding:
        burial_depth: 
          constant: 0
        release_start: 2016-09-21 19:30:00
        quantity: 1
        strategy: 
          random:
            bbox: "39400,16800 40600,17800"
            seed: 42
            nlocations: 300
    - name: pop2_noprob
      particle_type: sand
      characteristics:
        grain_size: 0.00025 
        density: 2650.0  
      tracer_methods:
        vanwesten:
          flow_field_name: 
            - bed_load_velocity
            - suspended_velocity
      transport_probability: no_probability # Options: no_probability, stochastic_transport, reduced_velocity
      seeding:
        burial_depth: 
          constant: 0
        release_start: 2016-09-21 19:30:00
        quantity: 1
        strategy: 
          random:
            bbox: "39400,16800 40600,17800"
            seed: 42
            nlocations: 300
outputs:
  directory: ./results
  store_tracks: true
  save_interval: 15M
visualization:
  dashboard:
    enable: false
    update_interval: 1H
```

sgpearson17:
Hi Bart, after chatting with @jreyns this morning, we think we figured out my issue; I have applied the fixes in the above commits and added some new tests:
image
image
Now we no longer get "tunnelling" behaviour when `no_probability` is enabled (see above)

bartvanwesten:
Hi Stuart,

I ran your yaml on dev and on this branch and compared. I think there are three separate things here:

**The burial numbers.** On dev, your no_probability population ends the 2-day run at 8.6 m mean burial depth (max 13.8 m). On this branch both populations stay at 0 for the whole run. The bed in `inlet_sedtrails.nc` doesn't change in time, so all of that burial was the spatial leak from #333. That's the part this PR fixes.

**Why your re-run looks unchanged.** A trajectory plot can't show the fix: no_probability particles are never flagged buried (`update_status` sets `status_buried=False` for that mode, since d54509a), so burial never stops them and their tracks are the same with or without the fix — I checked, positions are bit-identical between dev and this branch. The difference only shows up in burial_depth itself, see panel (d) below.

**Buried-but-mobile.** Since the buried flag is forced off for no_probability but `update_burial_depth` keeps accumulating, you get particles metres "deep" that still move. I'd either drop the burial bookkeeping for that mode (as in your branch) or apply the mixing-depth check in all modes. Separate PR either way — do you have a preference?

**The trapping at the island** is indeed a separate problem: the land-boundary behaviour from #270. Without island polygons the FM file gives us points without connectivity, the triangulation runs straight across the island, and particles get velocities interpolated from both sides pushing them onto the edge. That also produced the big bed-level differences feeding the burial in point 1. With polygons those triangles are removed, which is why it goes away.

PR 440 verification: trajectories identical on both branches, burial depth fixed

sgpearson17:
Ok thanks Bart! Let's discuss tomorrow, as I want to make sure I didn't try to fix something the wrong way... may need to dig back through my last commits that I made just before you posted this...

sgpearson17 (approved):
After discussion with @bartvanwesten:
The current version works as expected, see images below
image
image
I also added a (skipped by default) integration test to check burial stuff which generates the above two plots.

We will address the more general case when we solve issue #446

@jreyns , if you agree, then we can merge this PR and continue.

